PR #2, opened 2021-03-18
Added Dockerfile and docker-compose.yml for prod and dev

will0684:
Added a Dockerfile and docker-compose configuration to the project to support development and production deployment.

For local development simply run `docker-compose up dev` and docker will build the image, run it and map your volumes and ports for local development supporting hot-reload.


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+FROM node:current-alpine AS production
+ENV NODE_ENV=production
+WORKDIR /app
+COPY --from=build /build/package*.json ./
+COPY --from=build /build/.next ./.next
+COPY --from=build /build/public ./public
+RUN npm install next
+
+EXPOSE 3000
+CMD npm run start
```
pcraig3, line 22: So, ran it locally and it all works, but I'm wondering if you can run me through the steps here. 
When I've set up Dockerfiles previously they've been less involved than this.

For example: https://github.com/pcraig3/vacs/blob/main/Dockerfile

<details>
<summary>click to open</summary>

```
# use node alpine
FROM node:12-alpine

# set env var "PORT" to 3000
ENV PORT 3000

# Create app directory
RUN mkdir -p /app
WORKDIR /app

# Copy source files
COPY package*.json /app/
COPY . /app/

# Install  dependencies
RUN npm install --production --silent

# Building app
RUN npm run build

EXPOSE 3000
CMD ["npm", "run", "start:production"]
```
</details>

I guess my question is why do we say `FROM` 3 times?

In the repo I linked to, I also have a `.dockerignore` file so that we don't copy everything.

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+  npm:
+    <<: *common-props
+    entrypoint: npm
+  dev:
+    <<: *common-props
+    ports:
+      - "3000:3000"
+    command: npm run dev
+volumes:
+  node_modules:
```
pcraig3, line 22: Do we need a docker-compose file right at the start? I am imagining we Helm uses it but I don't know for sure.

package-lock.json
```diff
@@ -1,8 +1,2090 @@
 {
   "name": "alpha-site",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
```
pcraig3, line 5: Is something going on here? I didn't expect a 2000+ line change to the package-lock.json for adding a Dockerfile.

pcraig3 (commented):
Hello, left a few comments just to clarify some stuff. If you don't know the answers @will0684, feel free to reach out to @Moro-Code.

Dockerfile
pcraig3: Just to be clear, I'm not requesting changes, just wondering what's up.

will0684: Sure! This is kind of new to me as well but from my understanding those multiple FROM statements are to make a multi-stage build that reduces the final image size by only copying the most necessary files through the stages and leaving what's not needed behind. So in this case the node_modules are left out of the image. Its all based on this example here:

https://dev.to/chrsgrrtt/dockerising-a-next-js-project-1ck5

docker-compose.yml
will0684: This is also from the example I linked in the previous comment. I would say that we don't _need_ it but I thought it looked like a convenient setup that just meant devs don't have to pass in options to docker to map the volumes or expose the ports.

package-lock.json
will0684: I believe when I merged in main and fixed the conflict it may have restructured the file. I am not sure why the diff is so large myself.

will0684: I ended up just copying the lock from main. Pushing up now.

Dockerfile
Moro-Code: This is actually very smart. If nothing new is installed the npm install step will be skipped and go straight to building. I like this well done

docker-compose.yml
Moro-Code: This is interesting why the seperate npm container ?

Moro-Code (dismissed):
Well done I like the clever usage of layers in the Dockerfile

docker-compose.yml
will0684: I thought I understood the reason for it but I don't believe I do after some experimentation. So I removed it in the last commit.

pcraig3 (approved):

PR #3, opened 2021-03-18
Home page Markup

EphraimAssi:
I did a markup for the home page following the Wire frame design.

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -30,10 +30,7 @@ export default function Home() {
         <section>
           <div>
             <img src="#" alt="serviceCanada.jpg"></img>
-            <h1>
-              Servica Canada Labs
-              Make government work for you
-            </h1>
+            <h1>Servica Canada Labs Make government work for you</h1>
             <span>
               <button>Become a tester</button>
               <button>Start testing</button>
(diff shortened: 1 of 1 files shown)


pages/index.js
```diff
@@ -3,63 +3,117 @@ import styles from '../styles/Home.module.css'
...
-          Welcome to <a href="https://nextjs.org">Next.js!</a>
-        </h1>
+      <header>
+        <figure>
+          <img src="#" alt="logo.jpg"></img>
+          <figcaption>
+            Governement of Canada
+            Gouvernement du Canada
+          </figcaption>
+        </figure>
```
pcraig3, line 19: I would say double-check what Canada.ca uses for their header. We probably don't want a figure here, they are using `<div>`s on Canada.ca

Eg: https://design.canada.ca/mandatory-templates/home-page.html#specifications
"Working examples" links to the main canada.ca homepage: https://www.canada.ca/en.html

For the header, it looks like they are using 

```
<div>
    <a href="#" >
	<img src="..." alt="Government of Canada" property="logo">
        <span lang="fr">Government of Canada</span>
</div>
```

pages/index.js
```diff
@@ -3,63 +3,117 @@ import styles from '../styles/Home.module.css'
...
 
-        <p className={styles.description}>
-          Get started by editing{' '}
-          <code className={styles.code}>pages/index.js</code>
-        </p>
+        <section>
```
pcraig3, line 21: Let's do `<div>` here. Usually you want to use `<section>` in the page body, not the header element.

pages/index.js
```diff
@@ -3,63 +3,117 @@ import styles from '../styles/Home.module.css'
...
-            <p>Learn about Next.js in an interactive course with quizzes!</p>
-          </a>
+        <section>
+          <figure>
+            <img src="#" alt="serviceCanada.jpg"></img>
+            <h1>
+              Servica Canada Labs
+              Make government work for you
+            </h1>
+            <figcaption>
```
pcraig3, line 37: Let's make this a `<div>` too.

pages/index.js
```diff
@@ -3,63 +3,117 @@ import styles from '../styles/Home.module.css'
...
-        <a
-          href="https://vercel.com?utm_source=create-next-app&utm_medium=default-template&utm_campaign=create-next-app"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Powered by{' '}
-          <img src="/vercel.svg" alt="Vercel Logo" className={styles.logo} />
-        </a>
+      <footer>
+        <figure>
```
pcraig3, line 92: This is probably fine as a `<div>`.

pcraig3 (changes requested):
Looks pretty good overall (not beautiful, but mostly we're all right here).

If you're wondering how to code something, canada.ca has most of the stuff you're looking for. Also there's the [Canada.ca Design System](https://www.canada.ca/en/government/about/design-system.html) which has details about a lot of this stuff.

pages/index.js
EphraimAssi: Perfect I will make those changes

pcraig3 (approved):
Thanks @EphraimAssi, this will cause a conflict with #5 (or that one with this one, depending on which is merged first), but should be easy enough to overcome.

PR #4, opened 2021-03-18
added PR template

will0684:
Added a template for PRs that should cover our needs.


.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
...
+## Acceptance Criteria
+
+Get the acceptance criteria for the story and/or task you are working on and put it all here.
+
+## Test Instructions
+
+1. Steps on...
+1. how to test...
+1. or inspect this...
+1. feature manually.
```
pcraig3, line 16: This is probably a 1., 2., 3., 4. list, eh?

package-lock.json
```diff
@@ -1,8 +1,3231 @@
 {
   "name": "alpha-site",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
```
pcraig3, line 6: Not sure why this keeps happening to your PRs, @will0684. The actual PR template looks fine, but we definitely don't want to include 3000+ lines in the package-lock file as we didn't add any new packages here.

pcraig3 (changes requested):
Made a couple comments. Mostly, we should remove the package-lock stuff.

.github/pull_request_template.md
will0684: You are correct, fixing now.

package-lock.json
will0684: I agree. I'll try to figure out what's going on here.

Moro-Code: could you be using node 14 as opposed to 12 ?

will0684: Yes I am using 15 it seems, I'll switch down to 12 for the time being.

.github/pull_request_template.md
will0684: Fixed

pcraig3 (approved):
Looks good! We could also look at putting in a french translation for this if any devs would like it.

PR #9, opened 2021-03-19
Button component

EphraimAssi:
I did a button component and unit testing following the servicePrototype as an example.
image


components/atoms/ActionButton.js
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Identify which button being clicked
+   */
+  id: PropTypes.string,
+
+  /**
+   * Inverted color styling on the buttons as an default option
+   */
+  invert: PropTypes.bool,
```
Moro-Code, line 65: do we need this prop now ?

components/atoms/ActionButton.js
```diff
@@ -0,0 +1,121 @@
...
+      } ${
+        props.disabled
+          ? "bg-gray-light text-gray-600 border border-gray-md "
+          : props.className
+      } text-sm px-4 focus:outline-none`}
+      onClick={props.onClick}
+      onMouseEnter={props.onHover}
+      onMouseLeave={props.onMouseLeave}
+      id={props.id}
+      data-cy={props.id}
```
Moro-Code, line 30: Would like to make data-cy a separate prop here. Minor thing but its cleaner if they are seperate

components/atoms/ActionButton.stories.js
```diff
@@ -0,0 +1,42 @@
...
+  title: "Components/Atoms/ActionButton",
+  component: ActionButton,
+};
+
+const Template = (args) => <ActionButton {...args} />;
+
+export const Primary = Template.bind({});
+export const FullyRounded = Template.bind({});
+export const Rounded = Template.bind({});
+export const WithCounter = Template.bind({});
```
Moro-Code, line 13: remove this story

components/atoms/ActionButton.stories.js
```diff
@@ -0,0 +1,42 @@
...
+export const StartAgain = Template.bind({});
```
Moro-Code, line 14: remove this story

components/atoms/ActionButton.js
```diff
@@ -0,0 +1,121 @@
...
+   */
+  dataCy: PropTypes.string,
+  /**
+   * Test id for e2e test
+   */
+  dataCyButton: PropTypes.string,
+  /**
+   * Enabled or disabled the button
+   */
+  disabled: PropTypes.bool,
```
Moro-Code, line 120: since we are passing in classNames now maybe we should have a prop to pass in styles for when the button is disabled

Moro-Code (changes requested):
Great work overall ! Some changes! Lets try adding a cypress test to make sure the buttons are on the page. 

Can we also make sure to have the same fonts and colours as in the mockup ? 

Screen Shot 2021-03-22 at 12 20 06 PM
Screen Shot 2021-03-22 at 12 18 53 PM


Screen Shot 2021-03-22 at 12 20 33 PM
Screen Shot 2021-03-22 at 12 20 21 PM

See tailwind documentation on how to add colours 
https://tailwindcss.com/docs/theme#colors

components/atoms/ActionButton.js
```diff
@@ -0,0 +1,121 @@
...
+  return (
+    <button
+      className={`flex justify-center content-center h-auto p-1 ${
+        props.fullyRounded
+          ? "rounded-full py-2 px-4"
+          : props.rounded
+          ? "rounded-md py-2 px-4"
+          : props.notRound
+          ? "rounded-sm"
+          : "rounded-md"
```
pcraig3, line 16: We only have 1 style of button in the mockups, one with a small border radius. Can we take out the "rounded" option until it shows up?

components/atoms/ActionButton.js
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Boolean flag that specifies the button should be completely rounded
+   */
+  fullyRounded: PropTypes.bool,
+
+  /**
+   * Boolean flag that specifies the button should be slightly rounded
+   */
+  rounded: PropTypes.bool,
```
pcraig3, line 75: I think let's get rid of the "rounded" options. See above comment.

components/atoms/ActionButton.stories.js
```diff
@@ -0,0 +1,42 @@
...
+export const Disabled = Template.bind({});
+
+Primary.args = {
```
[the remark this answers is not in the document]
pcraig3: Yeah, I also don't see the blue background in the storybook links. Do you know why that is?

pages/index.js
```diff
@@ -83,7 +97,14 @@ export default function Home() {
             Experiencing an issue with this product or have you spotted an
             error?
           </p>
-          <button>Report a problem</button>
+          <ActionButton
+            text={"Report a problem"}
+            rounded
+            className={
+              "bg-white text-blue-600 border border-blue-600 w-auto hover:bg-gray-200"
+            }
```
pcraig3, line 105: We pretty have 2 button styles so far, the blue and the alternate white with blue text. Can we make this the default for our use-case?

pcraig3 (commented):
Hi Ephraim, good work! I made some suggestions, mostly about reducing the props to pass in. Hope it all makes sense.

pcraig3 (changes requested):
Two more changes requested, based on looking at the Canada.ca buttons.

buttons


1. The [canada.ca buttons](https://design.canada.ca/common-design-patterns/buttons.html) have an "`:active`" state that's different than the focus + hover state. 'active' means that you are actually pressing the button. When you click on the canada.ca buttons, their colour changes (in addition to the focus ring).
2. The focus ring is more pronounced on the Canada.ca buttons because they are using outline offset. Our focus ring (especially for the blue button) is way less noticeable. I would say let's make the focus ring a bit more prominent for our buttons: either use the outline offset + border that canada.ca uses, or let's just make the border thicker.

Here's ours for comparison. Note how clicking on the buttons doesn't change the bg colour.

buttons2

pcraig3 (approved):
Looks good Ephraim, thanks!

PR #11, opened 2021-03-23
cypress-axe and jest-axe

will0684:
# Description

[Setup automated a11y testing](https://trello.com/c/JO8M1zcU/63-set-up-automated-accessibility-testing)

This pr includes [cypress-axe](https://github.com/component-driven/cypress-axe) and [jest-axe](https://github.com/nickcolley/jest-axe), configuration to support the reporting of a11y failures when running cypress tests and github workflow actions to support the testing in CI.

Additionally, a [custom NextJS Document](https://nextjs.org/docs/advanced-features/custom-document) has been created to allow for control of the html lang attribute and other html or body element attributes that the application needs.

## Acceptance Criteria

Unit and a11y tests can be automated in the CI.

## Test Instructions

1. `docker-compose up dev`
2. `npm run cypress:run`
3. See that tests pass
4. Remove the `lang` attribute on the `<Html />` element found in /pages/_document.js and save
5. `npm run cypress:run`
5. See that test fails

## Checklist

## Product and Sprint Backlog


cypress/integration/home.spec.js
```diff
@@ -1,12 +1,19 @@
...
         cy.visit("/");
+        cy.injectAxe();
     });
 
     it("landing page loaded", () =>{
         cy.url().should("contains", "/");
     })
+
+    it('Has no detectable a11y violations on load', () => {
+        cy.checkA11y(null, null, terminalLog)
```
pcraig3, line 17: Seems like a weird function signature with the two `null`s but it's in the example, so 🤷

cypress/plugins/index.js
```diff
@@ -16,7 +16,19 @@
...
+      log(message) {
+        console.log(message)
+  
+        return null
+      },
+      table(message) {
+        console.table(message)
+  
+        return null
+      }
```
pcraig3, line 32: is this being used? usually we try to get rid of console messages.

pages/_document.js
```diff
@@ -0,0 +1,24 @@
...
+// Read more here: https://nextjs.org/docs/advanced-features/custom-document
+class MyDocument extends Document {
+  static async getInitialProps(ctx) {
+    const initialProps = await Document.getInitialProps(ctx);
+    return { ...initialProps };
+  }
+
+  render() {
+    return (
+      <Html lang="en">
```
pcraig3, line 13: This is so that we can add the `lang="en"`, right?

pcraig3 (commented):
Left a few comments, but I think this is pretty good.

cypress/plugins/index.js
will0684: From what I understand, it is so that you can see the same error messages in console output from the Cypress testing UI.

https://github.com/component-driven/cypress-axe#standard-output

pages/_document.js
will0684: Yes to modify those attributes you have to create a custom Document "component" to override the Next default.

https://nextjs.org/docs/advanced-features/custom-document

cypress/integration/home.spec.js
will0684: Yeah not the prettiest signature...

cypress/plugins/index.js
Moro-Code: is there a need to return null ?

Moro-Code: is there a need to return null ?

Moro-Code (commented):
Looks great just had that one comment on returning null

cypress/plugins/index.js
will0684: It appears to be the typical implementation for the task event listener in Cypress

https://docs.cypress.io/api/commands/task#Usage

pcraig3 (approved):
Thanks @will0684, looks good!

PR #15, opened 2021-03-24
Canada.ca Typography

will0684:
# Description

[Canada.ca Typography](https://trello.com/c/Q6OnqXI0/88-canadaca-typography)

This PR includes configuration to the typography to match it with the style guide provided here:

https://design.canada.ca/styles/typography.html

Default Tailwind utility classes for fonts have been overridden to consist of a `font-display` for heading type elements and `font-body` for content elements. 

Additionally, a styles/globals.css global stylesheet has been added to match the explicitly defined design found in the Canada.ca style guide. This stylesheet sets all heading elements (h1 to h6) to use the 'Lato' font-family and body elements to use 'Noto sans' by default as well as prescribed font-sizes with a single breakpoint as dictated in the style guide. When working with custom elements you can still use the previously mentioned utility classes to set the font-family declaratively.

## Acceptance Criteria

Typography across the application should match the mandatory style guide provided by Canada.ca.

## Test Instructions

1. npm run dev
2. see that headings and content are using the correct fonts

## Help Requested

N/A

## Checklist

N/A

## Product and Sprint Backlog


styles/globals.css
```diff
@@ -0,0 +1,69 @@
+/* Canada.ca Typography */
+/* https://design.canada.ca/styles/typography.html */
+
+/* Font Families */
+h1, h2, h3, h4, h5, h6 {
+  font-family: 'Lato', 'system-ui', 'sans-serif';
```
pcraig3, line 6: We could say `bold` here and then not everywhere individually below.

pcraig3 (commented):
I don't know how to do this, but do you think you could get our components in storybook to use our default fonts as well?

If you can't figure it out, we can do it later, but have a look-see if there's any way to add global styling to them.

styles/globals.css
will0684: Yes of course 🤦

will0684: Fixed 👍

Moro-Code (changes requested):
Wondering if we should instead hook this up through tailwind. You can add fonts in a style sheet through an `@import`.

`fonts.css`
```css
@import url('https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,600;0,700;0,800;1,300;1,400;1,600;1,700;1,800&family=Raleway:wght@700&family=Roboto:ital,wght@0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&family=Pacifico&display=swap');
```
tailwind.config.js
```js
module.exports = {
  purge: [
    './pages/**/*.{js,ts,jsx,tsx}',
    './components/**/*.{js,ts,jsx,tsx}'
  ],
  darkMode: false, // or 'media' or 'class'
  theme: {
    extend: {},
    fontFamily: {
      'body': ["Open Sans"],
      'title': ["Raleway"],
      'handwriting': ["Pacifico"]
    }
  },
  variants: {
    extend: {},
  },
  plugins: [
    require('@tailwindcss/typography'),
  ],
}
```

will0684:
@Moro-Code @pcraig3 @EphraimAssi @nibivi77 @nghe0001 

Alright so some significant changes have been made to the content of the PR.

**Before:** I had created a global stylesheet for h1-h6 and body tags to always use the font-sizes found [here](https://design.canada.ca/styles/typography.html)

**Now:** The global stylesheet has been removed. Font sizes and the font family will need to be declared as a class on whatever element/component you are working with. To support this, there are now several custom font-size classes that can be used based on the same Canada.ca design guidelines. `font-h1` would be the class you use on your h1 elements, for example. These can be found in tailwind.config.js.

components/atoms/ActionButton.js
```diff
@@ -22,7 +22,7 @@ export function ActionButton(props) {
         props.disabled
           ? "bg-gray-light text-gray-600 border border-gray-md"
           : props.className
-      } text-sm px-4`}
+      } text-sm font-body font-bold`}
```
will0684, line 25: I removed the px-4 here as it is was a duplicate declaration (see line 9)

components/atoms/Banner.js
```diff
@@ -9,9 +9,13 @@ export const Banner = ({ siteTitle, headline }) => {
...
       <div className="md:container w-full md:w-3/4 xl:w-1/2 bg-dk-blue bg-opacity-90 text-white p-4">
-        <h1 className="text-4xl pt-4 pb-2 break-words">{siteTitle}</h1>
+        <h1 className="font-display font-bold text-h1 pt-4 pb-2 break-words">
+          {siteTitle}
+        </h1>
         <hr className="border-2 border-hr-red-bar bg-hr-red-bar bg-opacity-90 border-opacity-90 w-3/4" />
-        <p className="text-xl py-4 break-words">{headline}</p>
+        <p className="font-body font-bold text-p py-4 break-words">
+          {headline}
+        </p>
```
pcraig3, line 18: Is there a way to make the `h1`s and the `p`s take a default class? Adding a class to every single paragraph and header is something we can just avoid if we can set a global default. I saw you did something like this with your Next personal project, right?

will0684: Thank you this is a great call and you are totally correct. I will implement that now.

PR #21, opened 2021-04-01
Default layout

EphraimAssi:
# Description

[Default Layout](https://trello.com/c/yLIvs2oL/89-default-layout)

Make a page Layout based on the mock-up and Canada.ca. 
image

## Acceptance Criteria
- Create a layout the contains the header and the footer element.
- Creating a container element to wrap components in the correct margin
- Must adjust to different screen sizes

## Test Instructions

Test the layout of the page at different screen sizes.

## Help Requested

-I had to make multiple containers for the layout to work well, so we'll have to modify the footer in the layout later also.

## Checklist

## Product and Sprint Backlog


components/organisms/Layout.js
```diff
@@ -0,0 +1,98 @@
+import PropTypes from "prop-types";
+import Head from "next/head";
+import { ActionButton } from "../atoms/ActionButton";
+import { SearchBar } from "../atoms/SearchBar";
+import { useState } from "react";
+
+export const Layout = ({ banner, children }) => {
+  //State for Search input
+  const [searchText, setSearchText] = useState(null);
```
Moro-Code, line 9: where is this even used what is the point of this state

components/organisms/Layout.js
```diff
@@ -0,0 +1,98 @@
...
+          <img src="#" alt="canada.jpg"></img>
+        </footer>
+      </div>
+    </>
+  );
+};
+
+Layout.propTypes = {
+  // Optional banner element outside default margins
+  banner: PropTypes.element,
```
Moro-Code, line 91: Why? This means I have to pass jsx in the props, I should be able to just pass bannerText and bannerTitle and if thise two are provided the Layout component should generate the banner

components/organisms/Layout.js
```diff
@@ -0,0 +1,98 @@
...
+  //State for Search input
+  const [searchText, setSearchText] = useState(null);
+
+  //Handler for search input
+  function handleInputChange(e) {
+    setSearchText(e.target.value);
+  }
+
+  return (
+    <>
```
Moro-Code, line 17: Why is this a fragment, Should just be a main tag

components/atoms/Banner.js
```diff
@@ -7,13 +7,17 @@ import PropTypes from "prop-types";
...
 export const Banner = ({ siteTitle, headline }) => {
   return (
-    <div title="Banner image." className="bg-banner-img py-8 md:px-8">
-      <div className="md:container w-full md:w-3/4 xl:w-1/2 bg-dk-blue bg-opacity-90 text-white p-4">
-        <h1 className="text-h1-xl font-medium pt-4 pb-2 break-words">
-          {siteTitle}
-        </h1>
-        <hr className="border-2 border-hr-red-bar bg-hr-red-bar bg-opacity-90 border-opacity-90 w-3/4" />
-        <p className="text-base font-normal py-4 break-words">{headline}</p>
+    <div title="Home banner" className="bg-banner-img py-8">
```
Moro-Code, line 10: div soup, you can achieve the same result with less divs

components/molecules/EntryBanner.js
```diff
@@ -0,0 +1,27 @@
+import Banner from "../atoms/Banner";
+import { ActionButton } from "../atoms/ActionButton";
+
+export const EntryBanner = () => {
```
Moro-Code, line 4: A component where text is hard coded is not a component at all. Hard coded text in a component makes i18n very difficult, all text should come from props or there should not be a component at all. I am of the opinion that this component should not exist and instead this should be done in the homepage

components/molecules/EntryBanner.js
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <section>
+      <Banner
+        siteTitle="Service Canada Labs"
+        headline="Make government work for you."
+      />
+      <div className="bg-gray-light-200 py-6">
+        <div className={"flex container xxs:mx-auto lg:px-6 xxl:mx-auto"}>
+          <ActionButton
```
Moro-Code, line 13: Again when these buttons are clicked... how are we supposed to respond to them if there is no callback ?

components/organisms/Layout.js
```diff
@@ -0,0 +1,98 @@
...
+      <Head>
+        <title>Alpha Site</title>
```
Moro-Code, line 19: no hard coding text in components

components/organisms/Layout.js
```diff
@@ -0,0 +1,98 @@
...
+  const [searchText, setSearchText] = useState(null);
+
+  //Handler for search input
+  function handleInputChange(e) {
+    setSearchText(e.target.value);
+  }
+
+  return (
+    <>
+      <Head>
```
Moro-Code, line 18: Need more here. If you're handling setting the head here such as the ability to set rich preview text and images for social media cards. Otherwise best to just remove and handle this in the respective pages

components/organisms/Layout.js
```diff
@@ -0,0 +1,98 @@
...
+                <img src="#" alt="Governement of Canada "></img>
+                <span lang="en">Government of Canada </span>
+                <span lang="fr">Gouvernment du Canada</span>
+              </a>
+            </div>
+
+            <div>
+              <a href="#">Français</a>
+              <SearchBar
+                placeholder={"Search Canada.ca"}
```
Moro-Code, line 39: no hardcoded text in Components for i18n

components/organisms/Layout.js
```diff
@@ -0,0 +1,98 @@
...
+              />
+            </div>
+
+            <nav>Menu</nav>
+          </div>
+          {banner}
+        </header>
+        <div>
+          <div className="flex container xxs:mx-auto lg:px-6 xxl:mx-auto mt-10">
+            <main>{children}</main>
```
Moro-Code, line 51: this should not be main

components/organisms/Layout.js
```diff
@@ -0,0 +1,98 @@
+import PropTypes from "prop-types";
```
Moro-Code, line 1: text not collapsing properly

Screen Shot 2021-04-01 at 10 32 41 AM

Moro-Code (changes requested):
Some changes needed. Good work

components/molecules/EntryBanner.js
pcraig3: These buttons are basically just links. They don't actually submit any data.

We should really use links here, although we can style them as buttons, or we can talk to Fred about changing how they look (better outcome).

components/organisms/Layout.js
pcraig3: Yeah, you pretty much want

```
<html>
  <head>
</head>
<body>
    <header> ... </header>
    <main> ... </main>
    <footer> ... </footer>
  </body>
</html>

File: components/molecules/EntryBanner.js (hunk 6 above)
Comment (pcraig3) [reply components/molecules/EntryBanner.js:13]: Actually, maybe we can leave them as buttons for now and just keep a backlog story about links that are buttons.

**Review by pcraig3 — CHANGES_REQUESTED** [review]
I have one more comment here, looking at the breakpoints we have defined. 

## Our "big screen" breakpoint

Our "big screen" breakpoint is set around `1024px`, vs the Canada.ca one, which is set around `992px`. It would be good to standardize around the Canada.ca one, although the mobile breakpoint stuff earlier is more important.

## Our "mobile" breakpoint(s)

By "mobile" I mean the breakpoint where our banner becomes full-bleed instead of floating on top of the picture.

| on canada.ca | in this PR |
|--------------|------------|
|    Screen Recording 2021-04-01 at 14 20 26     |    Screen Recording 2021-04-01 at 14 19 50   |
|  Once we hit the Canada.ca mobile breakpoint (`767px`), the margins are fixed. They don't ever shrink or snap down to a new size.       |  Once we hit our breakpoint (`767px` as well), we continue to snap the margin down when we hit smaller screen sizes. We should match Canada.ca and keep it consistent once we're below 767.   |

## Our 320px breakpoint

The smallest phone screen you really have to account for is at `320px`. Most phones are bigger, but older model iphones, or my old windows phone, would get as low as 320px. 

When I look at ours at 320px, the search bar doesn't fit on the screen anymore (but the Canada.ca one does).

Screen Shot 2021-04-01 at 14 30 36

We should make sure the search bar fits on the screen if we're at 320 px.

**Comment by EphraimAssi** [comment]
I left the Head in the Layout because I defined a meta for width to always be full screen in every page. Page doesn't fit on all screen sizes if I don't do that. The top of the header and the footer will be replaced so I also didn't  think it was worth putting props for the  text in there. I put some for the banner. And the search bar is being adjusted for different screen sizes in the header that Kevin is doing so I didn't change it also.

File: components/organisms/Layout.js [hunk 12]
```diff
@@ -0,0 +1,117 @@
...
+  bannerBecomeTesterText,
+}) => {
+  return (
+    <div>
+      <Head>
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+      </Head>
+      <div className="overflow-x-hidden">
+        <header>
+          <div className="lg:container xxs:px-6 xs:px-0 xs:mx-6 lg:mx-auto lg:px-6 xxl:mx-auto">
```
pcraig3, line 21: Two points here.


1. At the "xs" breakpoint, we have "6" horizontal margin and "0" padding, and then once we get to "xxs", we add "6" horizontal padding and the margin disappears. It seems like we could delete `xxs:px-6` and then use `xs:px-0 xxs:mx-6`.
2. The `lg:container` class sets breakpoints but there are still 7 classes that we're adding to 4 different body elements. Can we roll them up into one class by combining them so we have less copy+paste?

pcraig3 (changes requested):

styles/globals.css
```diff
@@ -37,39 +37,25 @@
...
-  @screen xl {
-    .layout-container {
-      max-width: theme("screens.xl");
-    }
-  }
-
-  @screen xxl {
-    .layout-container {
-      max-width: theme("screens.xxl");
+      @apply container mx-auto px-6;
```
pcraig3: This is awesome, good work!

pcraig3 (approved):
Thanks @EphraimAssi, let's merge this bad boy.

PR #25, opened 2021-04-08
Appservice

blai0264:
Updated context script
Added TF files for appService, trafficManager, and webApplicationGateway directories
Updated variables and outputs


andr0272 (approved):
Nice work.

PR #28, opened 2021-04-14
Feature/#127 alpha banner

nibivi77:
# Description

[MVP Alpha header banner] https://trello.com/b/ZqWtJSyh/alpha-site-board

This is a visual feature. It is the Alpha banner at the top of every page.
image

## Acceptance Criteria

N/A

## Test Instructions

1. Go to branch name feature/#127-AlphaBanner
2. npm run dev
3. See the banner at the top of the page.
4. Test multiple screen size width.

## Help Requested

Feedback is always appreciated.

## Checklist

- [ x] Strings use placeholders for translation (No hard coded strings)
- [ x] Unit tests have been added/updated

## Product and Sprint Backlog

[MVP Alpha header banner] https://trello.com/b/ZqWtJSyh/alpha-site-board


components/atoms/AlphaBanner.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+/**
+ * Displays the Alpha AlphaBanner on the page
+ */
+
+export const AlphaBanner = ({ text }) => {
+  return (
+    <div className="bg-gray-100">
```
Moro-Code, line 10: banner too thick, in the mockups on figma banner is 55px in height

components/atoms/AlphaBanner.js
```diff
@@ -0,0 +1,32 @@
...
+      <div className="lg:container xxs:mx-0 xxs:px-0 lg:mx-auto lg:px-0 xxl:mx-auto">
+        <div className="xxs:w-screen lg:w-2/3 xl:w-1/2 p-4">
+          <div className="flex">
+            <span className="max-h-9 uppercase border-2 border-black p-1 tracking-widest ml-2">
```
Moro-Code, line 14: There is no letter spacing in the mockups 

This is what you have:

Screen Shot 2021-04-15 at 12 16 25 PM

vs:

Screen Shot 2021-04-15 at 12 17 12 PM

font size should be 16px and there should be 10px or so padding

components/atoms/AlphaBanner.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Displays the Alpha AlphaBanner on the page
+ */
+
+export const AlphaBanner = ({ text }) => {
+  return (
+    <div className="bg-gray-100">
+      <div className="lg:container xxs:mx-0 xxs:px-0 lg:mx-auto lg:px-0 xxl:mx-auto">
+        <div className="xxs:w-screen lg:w-2/3 xl:w-1/2 p-4">
```
Moro-Code, line 12: feel like all these divs are unnecessary. Seems like this can be accomplished using two divs or so with the `layout-container` css class

Moro-Code (changes requested):

pcraig3 (dismissed):
Looks pretty good right now, the main thing I would say is that we should try and skim some of the fat off. Since this will be at the top of every page (including mobile), we want to reduce the footprint where possible.

I grabbed some comparison screenshots from [the COVID 19 Benefits site](https://covid-benefits.alpha.canada.ca/en/start). (note, you can click on the image to open them each in fullscreen mode and then swap between the tabs -- it's pretty tough to see the difference otherwise.)

| C19 Benefits | In this PR |
|---------|--------|
|  Screen Shot 2021-04-15 at 13 13 49     |    Screen Shot 2021-04-15 at 13 13 29  |

- Our font is a bit bigger and there is more padding inside of our tag.
- Also, on mobile, the font size goes down a little on the C-19 one. Let's do the same thing

Looking good though, nearly done.

components/atoms/AlphaBanner.js
```diff
@@ -0,0 +1,32 @@
...
+    </div>
+  );
+};
+
+AlphaBanner.propTypes = {
+  /**
+   * text area in the AlphaBanner
+   */
+  text: PropTypes.string.isRequired,
+};
```
pcraig3, line 30: Also, I feel like a better interface for this component would be not to call it an `AlphaBanner` specifically, but a `PhaseBanner`, to pass in the word Alpha as a prop (we can make it the default), and then have it take `children`.

More like this

```
<PhaseBanner phase="Alpha" >This site will change because it's an alpha.</PhaseBanner>
```

components/atoms/AlphaBanner.js
nibivi77: With tailwind, the closest I can get is 56px.

components/atoms/AlphaBanner.js
nibivi77: Without the letter spacing, the Alpha rectangle does not look the same. Probably due to the different font-family.

As for the font-size, the COVID 19 Benefit Finder page has a font-size of 14.76px. What should I follow? 

For the padding, I tried following the COVID 19 Benefit Finder as close as possible with tailwind.

components/atoms/AlphaBanner.js
nibivi77: With the new commit, I managed to remove one div. Not sure how to remove more :s

components/atoms/AlphaBanner.js
nibivi77: Font-size has been fixed to go small on mobile view.
padding inside the Alpha tag has been changed to match as close as possible the COVID 19 Benefit Finder. With Tailwind it is hard to get the exact pixel.
The AlphaBanner has been renamed to PhaseBanner. The requested change has been made to pass in children as the text and the phase as the title.

components/atoms/AlphaBanner.js
will0684: Thanks for your work so far Ben, made some additional changes to bring it a bit closer to the mockups here: https://www.figma.com/file/t5uDkPa86nsS3ssFESX2Em/BDM-DECD-Alpha-Site-Sprint-3?node-id=13%3A729

components/atoms/AlphaBanner.js
will0684: Changes addressed

components/atoms/PhaseBanner.js
```diff
@@ -0,0 +1,36 @@
...
+import PropTypes from "prop-types";
+
+/**
+ * Displays the PhaseBanner on the page
+ */
+
+export const PhaseBanner = ({ phase, children }) => {
+  return (
+    <div className="bg-gray-100 sm:max-h-14">
+      <div className="xxs:w-screen xxs:mx-0 xxs:px-0 lg:container  lg:mx-auto lg:px-0 xxl:mx-auto">
```
Moro-Code, line 11: hmmm this looks like it could just use the layout-container class

components/organisms/Layout.js
```diff
@@ -23,7 +26,18 @@ export const Layout = ({
   return (
     <div className="overflow-x-hidden">
       <header>
-        <div className="layout-container ">
+        {phaseText ? (
+          <PhaseBanner phase={phaseTitle}>{phaseText}</PhaseBanner>
```
Moro-Code, line 30: Think this is pretty static across all pages, not sure if we need to pass it in for each page

Moro-Code (commented):
All looks good some minor things.

Moro-Code (approved):
Looks good now

PR #29, opened 2021-04-14
Storybook i18n - Language Toggle

will0684:
# Description

No task for this on the board but I thought I would give it a try and see if I could get it working by end of day.

This PR adds a language toggle within the Storybook UI that on selection of a language will change a global value that Storybook can see and reference within individual Stories along with setting the locale of the i18n provider. This means translation can be done within the Stories through their args simultaneously with any strings that are fetched with `useTranslation()` that may not be managed through Story args. 

image

You can read about Storybook Globals [here](https://storybook.js.org/docs/react/essentials/toolbars-and-globals#advanced-usage). 

A Decorator is also used to ensure that every Story is wrapped in the i18n provider. You can read about Decorators [here](https://storybook.js.org/docs/react/writing-stories/decorators). This means that child components with strings that are not managed by the parent story's args/props have their strings appear as translated for the selected locale in Storybook as long as they are configured within the component with the `useTranslation()` hook.

Strings passed into Story args can be translated by referencing the global locale and using a [switch case](https://storybook.js.org/docs/react/essentials/toolbars-and-globals#consuming-globals-from-within-a-story) or whatever you prefer.

## Acceptance Criteria

NA

## Test Instructions

1. `npm run storybook`
2. Go to Organisms/Layout/NoBanner or WithBanner
3. Toggle language
4. See language changes

## Help Requested

- Anything you think needs work

## Checklist

## Product and Sprint Backlog


Moro-Code (approved):
This is super smart !! I actually didn't know storybooks had this feature

PR #33, opened 2021-04-19
Feature/#128 local nav

EphraimAssi:
# Description

[Local Nav](https://trello.com/c/zlqgf5hp/128-128-local-navigation-no-sign-up-button)

Do a local navigation bar following the Figma design.
image
image
image

## Acceptance Criteria

Needs to adjust to different screen sizes and be a dropdown on mobile screens.
Skip the sign-up button

## Test Instructions

1. Test the menu on different screen sizes
2. Dropdown needs to be full screen on mobile
3. Check the hover state

## Checklist

## Product and Sprint Backlog


Moro-Code (changes requested):
Good work ! Some changes needed 

Along with the changes at the bottom the close button should probably be a little further away from the top right hand corner 

Screen Shot 2021-04-19 at 3 08 50 PM

components/organisms/Layout.js
```diff
@@ -32,9 +38,50 @@ export const Layout = ({
...
+            title="Menu"
+            className="lg:container lg:mx-auto lg:px-6 lg:justify-end lg:flex"
+            data-cy="menu"
+          >
+            <div className="layout-container pb-4">
+              <button
+                onClick={onMenuClick}
+                id="menuButton"
+                className="text-h4 focus:outline-none focus:ring-2 focus:ring-black"
+              >
```
pcraig3, line 54: I think this should be properly hidden when not in use. Right now, it still takes up space and padding.

pcraig3 (changes requested):
Hi Ephraim!

There's one minor thing I noticed and then a bigger thing. 

- There's a bit too much space below the links on desktop mode: we should have equal space above and below the menu (although there will appear to be more space underneath once the breadcrumbs come thru)

## the menu interaction itself

There are 2 main ways to hide menus: modals or "reveal" interactions. The reveal ones are easier to do usually because the tab order doesn't need to be managed so much.  For example, here's one I did on the [COVID Alert Portal About page](https://covid-alert-portal.alpha.canada.ca/en/about/) (make your screen smaller to see it). 

When you do a fullscreen modal, you need to capture the focus, so that you can't tab off the modal, otherwise the keyboard interaction is all effed up. For example, if I open the menu on a small browser window, I can keep hitting tab until I start highlighting all the footer links, so we want to manage that. Also the "close" button isn't part of the tab order, so you can't reach it using a keyboard. 

I would say there are 2 options here:

1. Use a "reveal" instead of a modal: have the links open underneath the menu and disappear if the menu button is clicked again. Work with @stoltzf1 to include an "open links" mobile design because we don't have one currently 
2. Write some code to trap focus in the modal (including the "close" button). there's [an article here](https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex) and [a video here](https://www.youtube.com/watch?v=JS68faEUduk) that will be helpful. This is going mean writing more JS code and overall be more error prone so we should prefer the first option.

PR #36, opened 2021-04-20
Feature/#128 local nav

EphraimAssi:
# Description

[Local Nav](https://trello.com/c/zlqgf5hp/128-128-local-navigation-no-sign-up-button)

Do a local navigation bar following the Figma design.
image
image
image

## Acceptance Criteria

Needs to adjust to different screen sizes and be a dropdown on mobile screens.
Skip the sign-up button

## Test Instructions

1. Test the menu on different screen sizes
2. Dropdown needs to be full screen on mobile
3. Check the hover state

## Checklist

## Product and Sprint Backlog


pcraig3 (changes requested):
A couple of points here, but it's almost there. Looks really good, actually.

### menu button shouldn't move when clicked

Right now, the menu button moves side to side when clicked, but the menu button should remain in the same place. Also, the background colour doesn't cover the whole div so it looks a bit odd. Either the entire block should turn grey or we just keep everything white. Ask @stoltzf1 what we would prefer here.



### Let's add an "aria-expanded" attribute

Buttons that control open/close states should have an aria attribute that says true or false depending on the state it's in. There's [a smashing article](https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/) about it: it sums up by saying that a navigation menu button should: 

>  1. Identify itself as a button, not a link;
> 2. Identify the expanded or collapsed state of its corresponding menu (which, in strict terms, is just a list of links)

I checked out [what the bootstrap menus](https://getbootstrap.com/docs/4.0/components/navbar/) do, and they do this too, so we should be on solid ground here.

Screen Recording 2021-04-21 at 09 45 48

stoltzf1:
Agree the menu button should remain in same place. Let's actually keep the whole div white.

components/molecules/Menu.js
```diff
@@ -0,0 +1,81 @@
...
+        <button
+          id="menuButton"
+          onClick={onMenuClick}
+          className="text-h4 text-canada-footer-font focus:outline-none focus:ring-2 focus:ring-black"
+          aria-haspopup="true"
+          aria-expanded="false"
+        >
+          <span className="icon-menu" />
+          <span className="pl-3">{props.menuButtonTitle}</span>
+        </button>
```
pcraig3, line 37: I looked up if the HTML nesting matters for this, but I didn't see any definitive info. 

Eg, the button says `"aria-expanded"` but the menu it opens is in another div, not immediately after the button. 

Most of the examples I've seen are more like this.
```
<div>
  <button aria-expanded="true">menu</button>
  <ul><li>Item 1</li><li>Item 2</li></ul>
</div>
```

Whereas ours looks like this:
```
<div>
  <div>
    <button aria-expanded="true">menu</button>
  </div>
  <div>
    <ul><li>Item 1</li><li>Item 2</li></ul>
  </div>
</div>
```

Not too worried about it since I didn't see any specific documentation about this, but just thought I'd mention. 

Nothing to change here.

pcraig3 (commented):
Just one comment here. Made another one about "active" states in slack, so that's a small change to make.

pcraig3 (changes requested):
Last comment, I swear. I was looking at how other menus work, and found a good example on the CBC.

They are doing 2 things here: 

- On the button itself, they have `aria-expanded` set to true or false, depending on whether the menu is open or closed. (we are already doing this, so all good)
- They also have an attribute `aria-controls`, which points at the `id` of the element that is opening or closing. For us this would be the `menuDropdown` list, so let’s set aria controls to the ul id. 
- The last piece of the puzzle is the ul that opens and closes should update an `aria-hidden` attribute to true or false depending on it’s visible or not.

Once that’s all in, and it should be soon, we’re good. 

Adding a gif from the CBC menu here for reference. 

Screen Recording 2021-04-21 at 23 08 25

icomoon/demo.html
```diff
@@ -55,6 +55,20 @@ <h1 class="mvm mtn fgc1">Grid Size: 16</h1>
...
+            <fieldset class="fs0 size1of1 clearfix hidden-false">
+                <input type="text" readonly value="e900" class="unit size1of2" />
+                <input type="text" maxlength="1" readonly value="&#xe900;" class="unitRight size1of2 talign-right" />
+            </fieldset>
+            <div class="fs0 bshadow0 clearfix hidden-true">
+                <span class="unit pvs fgc1">liga: </span>
+                <input type="text" readonly value="" class="liga unitRight" />
+            </div>
+        </div>
     </div>
```
pcraig3, line 72: What's this demo file? Are we using this?

pcraig3 (approved):
I think it's good to go, but the test failed 👇

pcraig3 (approved):
Thanks @EphraimAssi!

Moro-Code (approved):
Approving now that tests have passed

PR #38, opened 2021-04-21
Feature/#132 experiment component

nibivi77:
# Description

[#132 Create an "experiment" component](https://trello.com/c/tAppcybk/132-132-create-an-experiment-component)

image

## Acceptance Criteria

 - The blue tag must be the same as the bottom border.
 - There are 3 data elements on the card: title, phase tag, description.

## Test Instructions

1. git checkout Feature/#132-ExperimentComponent 
2. npm run storybook
3. See Experiment component under atoms.
4. Make sure it matches the Experiment card on the mockups.

## Help Requested

Feedback on how to improve is always welcome.

## Checklist

## Product and Sprint Backlog

[#132 Create an "experiment" component]https://trello.com/c/tAppcybk/132-132-create-an-experiment-component


components/atoms/Experiment.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Displays an experiment card on the page
+ */
+
+export const Experiment = ({ title, tag, desc }) => {
+  return (
+    <div className="shadow-experimentShadow px-4 py-6 border-b-4 border-custom-blue-experimentBlue">
+      <h3 className="mb-4">{title}</h3>
+      <span className="py-2 px-1 text-white font-bold bg-custom-blue-experimentBlue rounded">
```
will0684, line 12: I would recommend using the class "inline-block" on the span so that the element's padding is reflected in the layout. Then you can use equal margin on the h3 and p tag and things won't shift around too much if the font-size or content were to change.

will0684 (commented):
I think it might be good to double-check in the standup if we should be following the font sizes in the mockups to the letter, because the font sizes used here are different.

will0684 (approved):
This looks good to me

Moro-Code (approved):
Looks good

Issue #42, opened 2021-04-22
Improvements to the Report a Problem widget


pcraig3:
It's pretty good, but there's a number of fixes that we should make

| desktop | mobile |
|---------|--------|
|  Our widget on [alphasite-main.dev.dts-stn.com](https://alphasite-main.dev.dts-stn.com/)    |  Widget on [Canada.ca Design System](https://design.canada.ca/common-design-patterns/report-problem.html)      |
|  <img width="312" alt="Screen Shot 2021-04-22 at 08 54 34" src="https://user-images.githubusercontent.com/2454380/115717998-b71c7380-a348-11eb-80c1-91b92e57d065.png">   |  <img width="350" alt="Screen Shot 2021-04-22 at 08 54 45" src="https://user-images.githubusercontent.com/2454380/115718007-b84da080-a348-11eb-8ae8-d9ba79a0c257.png">   |

### Report a problem `<summary>` element
- "Report a problem on this page"
- Text should be centred
- Text should be blue (like a link)
- Cursor should be a `pointer` when hovered  
- Active state should have a darker background
- They are using `border-style: outset` to make it "pop"

### Groups of checkboxes 
- Groups of related checkboxes / radio buttons should always be inside a `<fieldset>`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset
- The "check all the issues" text should be the `<legend>` for the fieldset (so, inside of the `<form>`)

### Checkboxes
- Let's make them 40 by 40px, not 44 by 44px
-  Remove 2px `top` and `left` CSS on input
- Set label:before `background` to `white`
- Set label:after `height` to 13px

### Font sizing
- 12px is way too small: let's set it to 16px again
- Line height should be `1.4` or `1.5`
- The "check all the issues" text at 18px is not enough contrast to 16px font, so increase it to 20 or 22px and increase the bottom margin a bit

### Width 
- Setting 300px seems tight to me. Can we set a `max-width` of 350-400 somewhere (maybe 380?) and then `width: 100%` so that it will size down as needed.

Maybe there's a bit more, but that will help a lot.

pcraig3:
The Life Journeys one has a pretty good `summary` implementation, here's a gif

![Screen Recording 2021-04-22 at 09 57 50](https://user-images.githubusercontent.com/2454380/115727273-7117dd80-a351-11eb-91ed-44c0d35e53ba.gif)

Moro-Code:
Solved by #43

PR #43, opened 2021-04-22
Fix/133 report a problem component

Moro-Code:
# Description

This PR aims to fix issues with the Report a Problem button outlined in #42 

### Report a problem `<summary>` element
- [x] "Report a problem on this page"
- [x] Text should be centred
- [x] Text should be blue (like a link)
- [x] Cursor should be a `pointer` when hovered  
- [x] Active state should have a darker background
- [x] They are using `border-style: outset` to make it "pop"

### Groups of checkboxes 
- [x] Groups of related checkboxes / radio buttons should always be inside a `<fieldset>`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset
- [x] The "check all the issues" text should be the `<legend>` for the fieldset (so, inside of the `<form>`)

### Checkboxes
- [x] Let's make them 40 by 40px, not 44 by 44px
- [x] Remove 2px `top` and `left` CSS on input
- [x]  Set label:before `background` to `white`
- [x] Set label:after `height` to 13px

### Font sizing
- [x] 12px is way too small: let's set it to 16px again
- [x] Line height should be `1.4` or `1.5`
- [x] The "check all the issues" text at 18px is not enough contrast to 16px font, so increase it to 20 or 22px and increase the bottom margin a bit

### Width 
- [x] Setting 300px seems tight to me. Can we set a `max-width` of 350-400 somewhere (maybe 380?) and then `width: 100%` so that it will size down as needed.

## Test Instructions

1. Check storybooks to see if all items have been addressed 

## Checklist

## Product and Sprint Backlog


pcraig3 (changes requested):
It's getting there. There are a few things that jump out at me.

- The font size on "report a problem" summary button seems too small. Can we make it 16px instead of 14?
- Let's make the background of the "questions" section the same colour and outline as the "summary" button. Right now the background is light blue for the questions, but on canada.ca both backgrounds are the same (or really close). You can see in the gif below.

### hover and active states 

| On Canada.ca | In this PR |
|---------|--------|
|      Screen Recording 2021-04-22 at 17 27 02    | Screen Recording 2021-04-22 at 17 26 09  |
|  when you click the button, you see an `:active` state, so the button gets darker than on hover. it also has a border surrounding it when the element is in `:focus`.   |    here we are missing a focus style and the active style. When you actually _click_ the button, you don't get a cool click interaction.    |

For the checkboxes, we are also missing focus and hover states.

| On Canada.ca | In this PR |
|---------|--------|
|   Screen Recording 2021-04-22 at 17 31 04   |  Screen Recording 2021-04-22 at 17 31 38     |
|  If you focus (tab + spacebar) [on the GDS ones](https://design-system.service.gov.uk/components/checkboxes/), you see a big yellow border. Same just after you click.  |  You can't see it, but the tab indicator disappears when you go into it. And when you click, there's no `:focus` styling.       |

Moro-Code:
@pcraig3 will implement changes from the font but for the background colour in figma the colour is light blue. Im not sure which one I should follow

stoltzf1:
Looks like this is the CSS from the Report a problem - Canada.ca design pattern (https://design.canada.ca/common-design-patterns/report-problem.html)

.well {
	min-height: 20px;
	padding: 19px;
	margin-bottom: 20px;
	background-color: #f5f5f5;
	border: 1px solid #e3e3e3;
	border-radius: 4px;
	-webkit-box-shadow: inset 0 1px 1px rgba(0,0,0,.05);
	box-shadow: inset 0 1px 1px rgba(0,0,0,.05);
}

pcraig3:
Okay @stoltzf1, so you think we should follow the Design System?

Moro-Code:
fixes done.. review when you can @pcraig3

pcraig3 (approved):
Looks good!

PR #47, opened 2021-04-26
implemented new SRE terragrunt model

blai0264:
Created new bootstrap script
Converted tfvars file to json
Removed variables.tf in each directory
Imported depot RG to terragrunt 
Imported remote storage account to terragrunt


ben851 (changes requested):
Please add an examples.json for the variables. Keep in mind the example should remove all sensitive content.

blai0264:
> Please add an examples.json for the variables. Keep in mind the example should remove all sensitive content.

updated.

ben851 (approved):

PR #52, opened 2021-05-02
Feature/137 report a problem api

Moro-Code:
# Description

[#137 Write Next JS API route for report a problem form submission](https://trello.com/c/q6UGarhN/137-137-write-next-js-api-route-for-report-a-problem-form-submission)

Wrote API route to take form submissions and submit to notify API to send an email to the specified email.

Note there are 4 new environment variables introduced in this PR. All of which have been added to the docker and helm setup as well as teamcity.

`NOTIFY_BASE_API_URL`: The base url for the notify API `https://api.notification.canada.ca`

`NOTIFY_API_KEY`: The notify API key used for authorization of request

`NOTIFY_REPORT_A_PROBLEM_EMAIL`: The email that the message should be sent too 

`NOTIFY_REPORT_A_PROBLEM_TEMPLATE_ID`: The notify template id to use for the email

## Acceptance Criteria

Done when a report a problem form submission results in sending of an email to an inbox with the details 

## Test Instructions

1. You'll need to have a notify account
2. Once you have one let me know and I'll invite you to the project 
3. Generate your api key
4. Create `.env.local` file with the env variables specified above 
5. Add your email to the safelist 
5. Run project and submit report a problem with at least one checkbox checked 
6. Submit form and see if you get an email

## Checklist

## Product and Sprint Backlog


Moro-Code:
> This pull request **introduces 1 alert** when merging [e4d780f](https://github.com/DTS-STN/Alpha-Site/commit/e4d780fec9f86d5edcc185b279a938b5a16abe2b) into [f16a307](https://github.com/DTS-STN/Alpha-Site/commit/f16a307810135b2265169ad7af9f83caef4af7b5) - [view on LGTM.com](https://lgtm.com/projects/g/DTS-STN/Alpha-Site/rev/pr-234698daa44d2f31a82ab62a276d90b42be324e4)
> 
> **new alerts:**
> 
>     * 1 for Unused variable, import, function or class

fixed by 97220d0

lib/reportAProblem/processData.js
```diff
@@ -0,0 +1,26 @@
+export const processData = (data) => {
+  // loop over keys in data and generate new object that aligns to notify body
+  return Object.keys(data).reduce((prevVal, currentVal) => {
+    // split value according to capitalized letters, make all lower case, and then join by underscore
+    let variableName = currentVal
+      .split(/(?=[A-Z])/)
+      .map((value) => value.toLowerCase())
+      .join("_");
```
pcraig3, line 8: Is this basically a translation between Javascript v. Python conventions? Would it be easier to use the `snake_case` variable names in the HTML and then never convert them?

Dockerfile
```diff
@@ -18,6 +18,10 @@ RUN npm run build
 
 FROM node:current-alpine AS production
 ENV NODE_ENV=production
+ENV NOTIFY_BASE_API_URL=https://api.notification.canada.ca
+ENV NOTIFY_API_KEY=""
+ENV NOTIFY_REPORT_A_PROBLEM_EMAIL=""
+ENV NOTIFY_REPORT_A_PROBLEM_TEMPLATE_ID=""
```
pcraig3, line 24: Should we add this to the README or something?

Currently, if you just try and submit without and env vars, you get a runtime error.

lib/reportAProblem/submitEmail.js
```diff
@@ -0,0 +1,50 @@
...
+          "The user did not report any issues with adaptive technologies accommodation.",
+        privacy_issues: "no",
+        privacy_issues_details: "The user did not report any privacy issues",
+        no_where_else_to_go: "no",
+        no_where_else_to_go_details:
+          "The user did not report not knowing where to provide feedback",
+        other: "no",
+        other_details:
+          "There is no unspecified issues that the user would like to report",
+        ...data,
```
pcraig3, line 44: I rewrote the `*_details` fields locally to pass in empty strings if nothing was submitted. We could just say "N/A" or something here since it just means that no details were passed in.

pcraig3 (commented):
A couple comments below, but this looks pretty good overall. 

I think my main worry is that the original email template is really verbose and it's hard to know what was actually submitted. I created a second email template that uses if statements to put headers in or not, so it only shows headings that where the checkbox was selected.

| before | after |
|--------|-------|
|  Screen Shot 2021-05-03 at 17 54 00   |   Screen Shot 2021-05-03 at 17 53 45  |

I think that's an obvious thing we can change. We should also change the "no input" message to something generic to all so that we can see at a glance if something was written vs it was left empty. Trying to scan the first email I sent quickly, it was really hard to tell.

### Node JS client

There's a node.js client offered by Notify: https://docs.notifications.service.gov.uk/node.html#set-up-the-client

Why not use it? I'm not super worried about it, just curious why you didn't want to. If the API call is updated in the future, our call to Notify could break and we wouldn't know because the error happens in the user's browser, not on the server. (The client doesn't help with this, it just makes it more likely that we would keep in sync with potential API changes).

Moro-Code:
>Why not use it? I'm not super worried about it, just curious why you didn't want to. If the API call is updated in the future, our call to Notify could break and we wouldn't know because the error happens in the user's browser, not on the server. (The client doesn't help with this, it just makes it more likely that we would keep in sync with potential API changes).

Not really super concerned about this. The API is super simple and we're making a single call to one end point. Adding another dependency doesn't really seem worth it. That's generally how I approach client libraries. Unless there is a lot of different endpoints I need to hit I usually just default to not using the client. I mean if they change the client then we'd have to change our code as well so same problem hypothetically

Dockerfile
Moro-Code: probably a good idea actually

lib/reportAProblem/processData.js
Moro-Code: that's probably a better idea yeah

Moro-Code:
>I think my main worry is that the original email template is really verbose and it's hard to know what was actually submitted. I created a second email template that uses if statements to put headers in or not, so it only shows headings where the checkbox was selected.

Much better template for sure. For some reason I thought when the variable was not present `incorrect_information_details` for example it would add spaces but I like that much better

lib/reportAProblem/submitEmail.js
Moro-Code: hmmmm but then you would have a series of N/A's with no headings ?

pcraig3 (approved):
Okay, let's ship it ✉️

PR #53, opened 2021-05-05
Add example env file with placeholder values

pcraig3:
# Description

Add example env file with placeholder values. You would still need to add a template ID and API key if you were planning to set up Notify.

There is no task for this on the board.


Moro-Code (approved):
LGTM

will0684 (approved):
Looks good thanks Paul

PR #66, opened 2021-05-18
Feature/filter experiments

Moro-Code:
# Description

[Cards should have a standardized size](https://trello.com/c/Qn0FsHsC/195-195-cards-should-have-a-standardized-size)
[Filter for the experiments](https://trello.com/c/JEvAH1TO/168-filter-for-the-experiments)

Desktop 
Screen Shot 2021-05-18 at 2 07 19 PM

Mobile 
Screen Shot 2021-05-18 at 2 07 39 PM

## Acceptance Criteria

168 - We have the labels for the filters determined, can add states as we go

195 - Cards here are differing sizes. Not very pretty. Lets make them all the same size according to figma.

## Test Instructions

1. Run tests
2. Run application and go to the experiments page

## Help Requested

N/A

## Checklist

## Product and Sprint Backlog


EphraimAssi (approved):
Looks good! The only thing I'm seeing is a console log for the filters, I don't know if we still need it.

Moro-Code:
> Looks good! The only thing I'm seeing is a console log for the filters, I don't know if we still need it.

Good catch !

components/molecules/Filter.js
```diff
@@ -0,0 +1,69 @@
...
+            label={label}
+            value={id}
+            name={id}
+            id={id}
+            dataTestId={id}
+            dataCy={id}
+            onChange={props.onChange}
+            checked={checked}
+            roundedFront={index === 0}
+            roundedBack={index === props.options.length - 1}
```
pcraig3, line 30: This is a neat way to do it.

pages/experiments.js
```diff
@@ -1,22 +1,67 @@
...
+      },
+      {
+        all: {
+          id: "all",
+          label: "All",
+          checked: filter === "all",
+        },
+      }
+    )
+  );
```
pcraig3, line 39: okay, so just humour me. 

It looks like whenever you click a particular filter, this loop re-executes every time, always coming up with the same result (eg, the same set of 4 filters). (ie, if you console.log on line 22 you see it). Should that be updating each time a filter is clicked? I get you redraw the experiments underneath, but is the point here that we want to set "checked" on whichever filter is tagged this way?

Seems like it's fine either way, was just curious.

pages/experiments.js
```diff
@@ -40,32 +92,37 @@ export const getStaticProps = async ({ locale }) => {
     {
       id: 1,
       title: "Lorem Ipsum",
```
pcraig3, line 94: Could you just add numbers to these maybe? Otherwise, you can't really see if they change because they all seem to say the same thing.

pcraig3 (changes requested):
A couple small changes here:

Screen Shot 2021-05-19 at 13 18 56

1. "Filtrer By" is spelled wrong. Let's use sentence case here, so "Filter by" (lowercase B)
2. Can we add some more margin under the "experiments" section? It's too close to the Report a problem button currently.

//

And then the last thing I would say is that when you're inside a form, pressing down "enter" usually submits the form (so in our case, it would select the checkbox). Since we don't have that logic, pressing "enter" doesn't work. (Not sure if it relies on there being a button in the form, but that could be it.)

Can you hook it up so that the "enter" keypress event is captured and selects the filter?

pages/experiments.js
Moro-Code: So we want to be able to add filters dynamically based on the content from the CMS this is the way to do that. Considering we're only going to have 3-4 elements maximum. This is almost a constant time approach

pages/experiments.js
Moro-Code: Sure thing

components/molecules/Experiment.js
```diff
@@ -0,0 +1,67 @@
...
+        "border-" + tagColours[props.tag] || "bg-gray-experiment"
+      }`}
+      data-testid={props.dataTestId}
+      data-cy={props.dataCy}
+    >
+      <h2 className="mb-2 text-p">{props.title}</h2>
+      <span
+        className={`inline-block py-2 px-2 uppercase font-body text-xxs text-white font-bold rounded ${
+          "bg-" + tagColours[props.tag] || "bg-gray-experiment"
+        }`}
```
pcraig3, line 26: The or `||` statements here don't ever evaluate to the second case. I tried creating a manual experiment here (without passing in a tag), and got `border-undefined` and `bg-undefined`.

Moro-Code: forgot to add brackets. Thanks for the catch

will0684 (dismissed):
Awesome work, nice robust testing. Looks good and is working as expected.

PR #67, opened 2021-05-18
List component

EphraimAssi:
# Description

[Create list component](https://trello.com/c/3WKW3Wce/186-186-create-list-component)

Create a list component that we can dynamically reference according to the data in the CMS.
image

## Acceptance Criteria

Should be able to pass in CSS classes in order to change sizing i.e. full width half width etc 

## Test Instructions

1. Test the component on different screen sizes
2. Test the component width, try changing the width

## Help Requested

After 4 items, the circles just stay dark blue. I'm not sure if we need to change this or not, any help appreciated!

## Checklist

## Product and Sprint Backlog


components/molecules/List.js
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ *  List component
+ */
+export function List(props) {
+  let opacity = 20;
+  return (
+    <ol className={props.className}>
+      {props.items.map((item, key) => {
+        opacity += 20;
```
Moro-Code, line 11: Opacity is gonna start at 40 this way. Assign a variable to store the JSX and then add 20 to the opacity after. Note if the opacity === 100 .... 20 should not be added

Moro-Code (changes requested):
Wow @EphraimAssi super great work! Code is very clean I'm impressed ! Just one thing... the opacity doesn't seem to be working and I left a very minor comment

components/molecules/List.js
EphraimAssi: The opacity in Figma design starts at 40 that's why I left it that way. Is it still better to put the JSX in a variable?

Moro-Code (changes requested):
Hey @EphraimAssi just need to fix the colours for the list item

What we want

Screen Shot 2021-05-20 at 9 11 21 AM

What we have

Screen Shot 2021-05-20 at 9 12 20 AM

will0684 (approved):
Looks good Ephraim nice work 👍

Moro-Code (approved):
LGTM!

PR #86, opened 2021-06-01
_app.js check for stored language, show splash if none

will0684:
# Description

[Hook up splash page](https://trello.com/c/D5Zys6KN/224-224-hook-up-splash-page)

This PR hooks up the splash page to the site and works as so:

When a user navigates to any page on the site, the top level `_app.js` component checks the client browser's localstorage for a `lang` property with some value. If the value is neither `en` or `fr`, the user is sent to the splash page to select their preferred language. Once selected, that value is stored under `lang` in localstorage as either `en` or `fr`.

Once the value is stored, on subsequent visits to any page on the site, the localstorage `lang` property will be checked and the user will be routed to the desired page and routed to their selected locale if required. 

For example if a user has selected French as their preferred language, and they try to access the url `alphasite.ca/experiments`, they will be automatically routed to `alphasite.ca/fr/experiments`. They can change their locale anytime with the language toggle on the page, which will also update the `lang` value in localstorage.

## Acceptance Criteria

* If user has never been on site for that browser ( no language set in local storage) and heads to the home page they should be redirected to the splash page to select their language

* Once a user selects their language it should be stored in local storage 

* If a user returns to the home page after navigating off ( has language selected in localstorage ) their language preference should be loaded in and they should be redirected to the appropriate language

* If a user switches languages by using the language toggle their new language preference should be stored

## Test Instructions

1. Clear your localstorage (at least the `lang` property)
2. Navigate to the site
3. Splash page should appear and select a language
4. You are taken to homepage after selecting language, appropriate locale is displayed with content in correct language
5. Toggle language, see that `lang` value is updating

## Help Requested

- Any improvements you think could be made

## Checklist

## Product and Sprint Backlog

### DIFF
--- a/components/organisms/Layout.js
+++ b/components/organisms/Layout.js
@@ -10,6 +10,15 @@ import { useTranslation } from "next-i18next";
 import { DateModified } from "../atoms/DateModified";
 import { SearchBar } from "../atoms/SearchBar";
 
+const setLanguage = (language) => {
+  if (typeof window !== "undefined") {
+    console.log(language);
+    language === "fr"
+      ? window.localStorage.setItem("lang", "fr")
+      : window.localStorage.setItem("lang", "en");
+  }
+};
+
 /**
  * Component which defines the layout of the page for all screen sizes
  */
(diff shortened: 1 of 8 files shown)


EphraimAssi (approved):
It works really well and fixes the lang attribute issue at the same time. Looking good!

Moro-Code (approved):
Ship it 🚀

PR #93, opened 2021-06-04
Skip button

EphraimAssi:
# Description

[Create a "Skip to Content" link in the layout for a11y](https://trello.com/c/1hxhUiQZ/231-231-create-a-skip-to-content-link-in-the-layout-for-a11y)

I made Skip to Content button accessible for keyboard users to skip straight to the main content without going through multiple things with tab.
image

## Acceptance Criteria

Skip to main content button should be the first thing reachable with tab and also should skip to the main title.

## Test Instructions

1. Test the button to see if it skips to the main title on each page (Home, Experiments, About)
2. It should always be invisible unless it's focused
3. It should always be centered horizontally no matter the screen size

## Help Requested

- I put the place under the alpha banner since like Canada.ca doesn't have an alpha banner but if anyone has suggestions for placement, I'm open to it!

## Checklist

## Product and Sprint Backlog

### DIFF
--- a/components/organisms/Layout.js
+++ b/components/organisms/Layout.js
@@ -10,6 +10,7 @@ import { useTranslation } from "next-i18next";
 import { DateModified } from "../atoms/DateModified";
 import { SearchBar } from "../atoms/SearchBar";
 import { Breadcrumb } from "../atoms/Breadcrumb";
+import { ActionButton } from "../atoms/ActionButton";
 
 const setLanguage = (language) => {
   language === "fr"
@@ -33,6 +34,15 @@ export const Layout = ({
 
   return (
     <div className="overflow-x-hidden">
+      <div className="skip-main">
+        <ActionButton
+          id="skipToMainContent"
+          text={t("skipToMainContentBtn")}
+          custom="bg-custom-blue-dark text-white py-1 px-2 hover:bg-gray-dark"
+          href="#pageMainTitle"
+          dataCyButton={"skip-Content"}
+        />
+      </div>
       <header>
         <PhaseBanner phase={t("Alpha")}>{t("alphaText")}</PhaseBanner>
         <div className="layout-container flex-col flex lg:flex lg:flex-row justify-between  mt-2">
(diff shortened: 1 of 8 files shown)


will0684 (changes requested):
Looking really good! One issue I'm seeing. When you click on the "Skip to content" button and then scroll back up on the same page and try to toggle the language, it won't let you change the language but instead keeps activating the skip to content.

pcraig3:
2 things, Ephraim.

1. The accessibility checklist:

On [the accessibility checklist](https://bati-itao.github.io/resources/a11ycheck-en.html), there's a list that says Focus is always visible (using keyboard). 
- When we use the skip link [on Canada.ca](https://www.canada.ca/en.html), it highlights the main header with.a focus ring
- On our page, we don't have a H1 focus style at all, so we should add one when it's focused.

2. Jordan is right, the french link stops working.

The french link seems like it literally updates the "href" on the page when you add a #hash to the end of a URL. We don't want the "French" link to update after clicking the skip link, so maybe we should try using an `<a>` tag here instead of the `<Link>` component for the language link: https://github.com/DTS-STN/Alpha-Site/blob/main/components/organisms/Layout.js#L45

components/organisms/Layout.js
```diff
@@ -33,6 +33,18 @@ export const Layout = ({
 
   return (
     <div className="overflow-x-hidden">
+      <nav className="skip-main">
+        <a
+          id="skipToMainContent"
+          className="bg-custom-blue-dark text-white py-1 px-2 hover:bg-gray-dark"
+          href="#pageMainTitle"
+          dataCyButton={"skip-Content"}
```
will0684, line 41: Getting a warning for this attribute 
image

will0684 (changes requested):
Nice work! One tiny thing and then it will be good to merge

pcraig3 (approved):
Tried it out again and it works great! Good work, @EphraimAssi! (Let's wait for Jordan to review too though)

will0684 (approved):

PR #101, opened 2021-06-09
Signup page - missing required label

var-kyle:
# Description

https://trello.com/c/zusIsasP

The consent checkbox at the bottom of the signup for is required, but was missing the required label. This PR adds that required label.

image

This is kind of a special circumstance since it's just an individual checkbox. In the case of the checkbox list (like the question about being part of a visible minority in the same form) the parent fieldset contains the required/optional label.  If I were to use the same logic for showing the required label as there is in the TextField component:

```
        {props.required ? (
          <b className="text-error-border-red">*</b>
        ) : undefined}{" "}
        {props.label}{" "}
        {props.required ? (
          <b className="text-error-border-red">(required)</b>
        ) : (
          <p className="inline text-form-input-gray text-sm">(optional)</p>
        )}
```
we would see the required/optional text in each item of the list, which is something we don't want.

Another option would be to change the design of this question so that it would follow the fieldset -> checkbox pattern.

## Acceptance Criteria

Show the required label for the consent checkbox in the user signup form.

## Test Instructions

1. Go to the signup form and scroll to the bottom
2. See that the required label is present in the question

## Checklist

## Product and Sprint Backlog


EphraimAssi (dismissed):
Looks good!

.idea/prettier.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
will0684, line 1: You mind adding this file and .vscode/launch.json to the .gitignore?

will0684 (approved):
LGTM! 👍

PR #110, opened 2021-06-11
Update pre-commit hook

var-kyle:
# Description

Using just Husky for the pre-commit hook required a `git add .` which doesn't work very well if you have only want to commit a couple of changed files since it would magically add all your edited files (both staged and unstaged) to the commit.

This PR updates the pre-commit hook to use `lint-staged` with Husky so that we can remove that `git add .`. lint-staged will only perform the formatting on files that are staged and won't randomly add unstaged files to your commit.

I am only doing this because I was loosing my mind the other day when my unstaged files were included in my commit.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npm run lint
+npm run pre-commit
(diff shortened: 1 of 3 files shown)


EphraimAssi (approved):
Awesome!

Moro-Code:
Hahahahaha the best solutions are born out of insanity 🤪

PR #123, opened 2021-06-16
Added script tags for staging

EphraimAssi:
# Description

[Add script tags to application document as required by documentation](https://trello.com/c/KIrWaU2e/246-246-add-script-tags-to-application-document-as-required-by-documentation)

I added script tags that are going to be used for analytics. For now, it looks like it's not collecting anything. I think it's related to the other analytics tasks. 

## Acceptance Criteria

It shouldn't crash and there should be an adobe script tag in the head of a page and also a script tag right before the end of the body tag with _satellite.pageBottom();

## Test Instructions

1. Run the code and see if the two scripts are there in the head and right before the end of body
2. It shouldn't crash saying the _satellite.pageBottom() function doesn't exist

## Help Requested

I read the documentation and put what they said, but some things weren't clear, so if you guys check the documentation, any feedback is welcomed!

## Checklist

## Product and Sprint Backlog


pages/_app.js
```diff
@@ -33,6 +33,7 @@ function MyApp({ Component, pageProps }) {
           content={process.env.NEXT_PUBLIC_BUILD_DATE}
         />
         <meta name="og:buildid" content={process.env.NEXT_PUBLIC_TC_BUILD} />
+        <script src="//assets.adobedtm.com/be5dfd287373/0127575cd23a/launch-913b1beddf7a-staging.min.js"></script>
```
var-kyle, line 36: We probably want to put the link in an env and then check to see if it exists before adding it to the page.
It should look something like:
```
...
{
  if (process.env.ADOBE_ANALYTICS_URL){
    <script src=${process.env.ADOBE_ANALYTICS_URL}></script>
  }
}
...
```

We don't want to collect analytics while developing, so locally we won't have that env, but it would exist in our Azure WebApp. This also means we don't have to make any code changes when going to prod.

pages/_document.js
```diff
@@ -15,6 +15,7 @@ class MyDocument extends Document {
         <body>
           <Main />
           <NextScript />
+          <script type="text/javascript">_satellite.pageBottom();</script>
```
var-kyle, line 18: Same idea as above, we'd want to check to see if we should add it.

var-kyle (changes requested):

pages/_app.js
will0684: This is a great point. @EphraimAssi I'm wondering if there is any mention of recommended practices in respect to this within the install documentation?

EphraimAssi: @will0684 They did mention whitelisting some domains that may block the scripts but nothing specific on the implementation.

will0684 (approved):
Tested, looks good to me 👍 nice work

var-kyle (changes requested):
Sorry, I forgot about it as well, but since this env will be on the client side, we need to add `NEXT_PUBLIC_` to the front of it.

var-kyle (approved):
Nice work.

PR #125, opened 2021-06-17
Incremental static regeneration

var-kyle:
# Description

[Set incremental static regeneration on pages](https://trello.com/c/OXGoo5wE)

[ISR](https://nextjs.org/docs/basic-features/data-fetching#incremental-static-regeneration) can be used to see updated data from the CMS instead of having to rebuild the whole site. This is done by invalidating the cached page after the specified period of time. Once the cache is invalidated, Nextjs will rebuild that page on the next request. It gets tricky deciding the timeout though, as it would depend on how often data in the CMS is updated. The example from Nextjs sets the timeout at 10 seconds, but that seems far too frequently, yet a day seems not frequently enough. I've opted to set it to invalidate every hour - of course this can be changed if we feel it necessary.

The other thing to note is that this is only on the experiments page. After the backlog refinement meeting, I think the decision was made not to let users use Strapi as a page builder, but to only let them determine content in certain areas (like the experiments). As far as I am aware, it's only the experiments page that gets data from Strapi.

## Acceptance Criteria

The experiments page updates the experiments from the CMS once an hour.

## Test Instructions

1. Navigate to the experiments page
2. Notice that there are x experiments
3. Navigate to Strapi (either local or the deployed instance, whichever your frontend is pointing to)
4. Updated an experiment, or add one, or even remove one
5. Wait an hour (you could actually just change the timeout. You just need update `revalidate` in `getStaticProps` on the experiments page)
6. Navigate to the experiments page again and you should notice the update you made in the CMS reflected there

## Help Requested

- I set the timeout to **1 hour** but this is still something up for discussion. It could also be put in an env so that prod/staging have different timeouts, but I hesitated about adding yet another env.

## Checklist

- [N/A] Strings use placeholders for translation (No hard coded strings)
- [N/A] Unit tests have been added/updated

## Product and Sprint Backlog

### DIFF
--- a/pages/experiments.js
+++ b/pages/experiments.js
@@ -130,5 +130,6 @@ export const getStaticProps = async ({ locale }) => {
       experimentData: data,
       filters,
     },
+    revalidate: 60 * 60 * 24, // invalidates the cache once a day
   };
 };


will0684 (approved):
Good stuff thanks Kyle

PR #134, opened 2021-06-23
Custom error page

var-kyle:
# Description

[Custom error page](https://trello.com/c/FWZ3U6Sy)

This page is kind of unique from the rest of the site. First, it uses server side rendering rather than static generation. This is because of two things:
- Reading the querystring is easier?
- Setting the status code

The first one could easily be just me not looking into how to read the querystring from the client side, but then we get into status codes. If an error page shows an HTTP 200, is it really an error? I'm not actually sure it really matters in our case since the only place that will be redirecting here is from the email validation endpoint - in which case, what status codes would you use for missing/invalid IDs?

Regardless of how the page is generated, I made it look like the 404 and 500 pages we have. The title and message are passed in via the querystring and they have default values of:
- Title: `The web site has reported an error. | Le site Web a signalé une erreur.`
- Message: `If the problem persists, report the problem. | Si le problème persiste, veuillez signaler le problème.`

And there is a link back to the home page.
image

## Acceptance Criteria

An error page is rendered with custom messaging.

## Test Instructions
This page is not currently hooked up anywhere, but you can still access it via `/error`

1. Navigate to `/error`
2. See an error page similar to the 404 or 500 error pages
3. Navigate to `/error?statusCode=418&errorTitle=I'm a little teapot&errorMessage=Short and stout&errorTitleFr=I don't know it in french&errorMessageFr=I'm sorry`
4. See that the heading and message are different than the first time landing on this page, and that the status code is not 200. 

## Checklist

## Product and Sprint Backlog


will0684 (changes requested):
Nice work! Perhaps we could have a simple unit test or two just to make sure it renders and perhaps to pass in props or test passing a query string?

will0684 (approved):
Nice job getting this working 🚀

PR #146, opened 2021-06-28
Fixed HTML validation errors

krischarbonneau:
# Description
[Fix HTML validation errors](https://trello.com/c/ebP2ENED)

The signup page was failing HTML validation because `p` tags were being used in spots that aren't allowed as per HTML5 standards. Based on current HTML standard, the the only elements allowed inside a `legend` tag are those in the Phrasing Content group (listed [here](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content)). To get around the errors showing up on the screenshot on Trello, I changed the `p` tags that were children of `legend` tags to be `cite` tags and removed the italic styling. I also did this with the `p` tag inside the `SelectField` component.

## Acceptance Criteria

Done when there the HTML validation errors are resolved.

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3.  Use your favorite validator extension to see that the errors are gone.

## Product and Sprint Backlog

### DIFF
--- a/components/atoms/SelectField.js
+++ b/components/atoms/SelectField.js
@@ -29,7 +29,9 @@ export function SelectField(props) {
         {props.required ? (
           <b className="text-error-border-red">{t("required")}</b>
         ) : (
-          <p className="inline text-form-input-gray text-sm">{t("optional")}</p>
+          <cite className="inline text-form-input-gray text-sm not-italic">
+            {t("optional")}
+          </cite>
         )}
       </label>
       {props.error ? <ErrorLabel message={props.error} /> : undefined}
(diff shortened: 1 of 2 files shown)


var-kyle (changes requested):
I don't think <cite> is an appropriate html element in this case, I would assume <span> would be fine.

krischarbonneau:
> I don't think `<cite>` is an appropriate html element in this case, I would assume `<span>` would be fine.

Fully agree, made the switch to `span`

var-kyle (approved):
Looks awesome, great work!

PR #151, opened 2021-06-30
Unsubscribe

var-kyle:
# Description

[Unsubscribe](https://trello.com/c/UCwRGy3G)

~~*This is still a work in progress*~~

An api endpoint will handle the form submission from /unsubscribe, which will generate a link and send it to the provided email address. Clicking the link will then perform the unsubscribe functionality, which is to remove all data from the record except the cuid.

## TODOs
Most of these can be done after we get actual designs, this PR can be looked at as just adding the functionality and the api endpoint, not the actual pages just yet.

- A confirmation page after a user unsubscribes
- Add a message on the /unsubscribe page after the user submits the form informing the user to check their email? Either that or have a second confirmation page - similar to how the signup/validate works
- ~~Unit testing~~
- E2E testing
- translations

## Acceptance Criteria

A user has the ability to unsubscribe from receiving invites to new experiments, or any other messages they might receive. There are two ways to do this: 
1. through a link provided in any email correspondence
2. the /unsubscribe page

After following the  /api/unsubscribe link from an email, the corresponding record in mongo contains only the cuid.

## Test Instructions

1. Follow the signup process, there is no need to do the validation
2. Navigate to the /unsubscribe page
3. Enter the email you used to sign up
4. Check the email with Unsubscribe in the subject and follow the provided link
5. Check your mongo database and see that the record is now only the cuid

## Checklist

## Product and Sprint Backlog


lib/users/unsubscribeUser.js
```diff
@@ -0,0 +1,35 @@
+import { Db } from "mongodb"; //lgtm [js/unused-local-variable]
+
+/**
+ * A function used to unsubscribe a user
+ * @param client {Db} - the connection object for the database
+ * @param cuid {string} - the cuid for the user
+ * @returns {Promise<Collection.insertWriteOpResult>}
+ */
+export async function unsubscribeUser(client, cuid) {
+  return await client.collection("users").findOneAndUpdate(
```
Moro-Code, line 10: Should deleteOne be used here instead. Couldn't help myself from taking a peek

var-kyle: I initially thought so too, but we now want to keep the cuid while removing any other data. Reasons can be found in [Design Doc 008](https://github.com/DTS-STN/Alpha-Site/wiki/Design-Doc-008-Unsubscribe).

pages/api/unsubscribe.js
```diff
@@ -0,0 +1,86 @@
...
+        process.env.MONGO_DB
+      );
+      try {
+        userObj = await getUserByEmail(conn.db, data.email);
+      } catch (e) {
+        console.log(e);
+        return res.redirect("/error");
+      }
+
+      if (userObj) {
```
nibivi77, line 23: If the email does not exist in the db, (I assume it returns null or undefined), would we want to return an error message? ``` else {
        return res.status(400).end("Email not found");
      } ```

pages/api/unsubscribe.js
```diff
@@ -0,0 +1,86 @@
...
+        return res.status(201).end("USER UNSUBSCRIBE EMAIL SENT");
+      }
+    }
+    return res.status(400).end("Missing email");
+  } else if (req.method === "GET") {
+    const id = req.query.id || "";
+    let unsubUserObj;
+
+    if (id) {
+      const conn = await connectToDatabase(
```
nibivi77, line 64: Same here for the connection to the db. Can it fail?

pages/unsubscribe.js
```diff
@@ -0,0 +1,238 @@
...
+
+  const [email, setEmail] = useState("");
+  const [emailError, setEmailError] = useState("");
+
+  const [errorBoxErrors, setErrorBoxErrors] = useState([]);
+  const [errorBoxText, setErrorBoxText] = useState("");
+
+  const handlerClearData = (e) => {
+    e.preventDefault();
+    setEmailError("");
```
nibivi77, line 45: I think this is a copy paste error. The handleClearData should `setEmail("")` and not `setEmailError("")`.

Unless we want to keep both so in this case just add `setEmail("")`

nibivi77 (commented):
This is really good and I am learning reading through the code!
I know that this is purely for the back-end side of it but I can't help myself to note some things down(**To be dealt with when we have a designed page**) :
- Error box crashes when clicking on the error.
- pages/unsubscribe.js has some hardcoded strings.

pages/api/unsubscribe.js
var-kyle: That is the right idea, but since this is an api route, that would just load a blank screen with the text `Email not found`. The `/error` route takes query string parameters for custom error messages and displays them in a similar fashion as the 404 and 500 error pages.

pages/unsubscribe.js
var-kyle: You are correct in that I did miss `setEmail("")`. In case there is an error, we want to clear it as well, so I'll keep `setEmailError("")` and add the other.

nibivi77 (approved):
Awesome work!

Issue #140, opened 2021-06-27
Error messages above the form don't focus the input (also lots of "undefined" classnames)


pcraig3:
The error messages in the validation errors warning box at the top should either move focus to the field label or the input itself. We should also be using anchor links instead of buttons for the warning messages.

So it should look more or less like this:

```html
<ul data-cy="error-box-items">
  <li>
    <a href="#email">Error 1: Must be a valid email</a> 
    <!-- OR -->
    <a href="#email-label">Error 1: Must be a valid email</a> 
  </li>
</ul>

<form>
  <label for="email" id="email-label">Email address</label>
  <div class="validation-message">Error 1: Must be a valid email</div>
  <input type="text" id="email" />
</form>
```

Right now, we are using buttons not links, the ids are the full sentence instead of something more meaningful (ids shouldn't really have whitespace), and there are a lot of undefined values in the classname there.

<img width="1229" alt="Screen Shot 2021-06-27 at 15 02 30" src="https://user-images.githubusercontent.com/2454380/123556335-b513ce00-d758-11eb-9b57-893849cecb4c.png">

will0684:
Addressed here: https://github.com/DTS-STN/Alpha-Site/pull/159

PR #153, opened 2021-07-02
Bug/344 error on click

nibivi77:
# Description

[Signup error click](https://trello.com/c/Uo1X4ZYo/344-344-fix-onclick-is-not-a-function-when-user-clicks-on-error-message)

When clicking on an item from the error summary for teh signup form, the app would crash - see Trello board item linked above.

## In This PR

Added the functionality to scroll the user to the desired error.
Added smooth scrolling behaviour to both, the sign up button, and the error summary items.

## Acceptance Criteria

N/A

## Test Instructions

1. `git checkout Bug/344-ErrorOnClick` and `npm run dev`
2.  Navigate to the sign up page.
3.  Scroll all the way down and press on the "Submit" button.
4. See the scrolling animation
5. Press on all 4 of the error items to see where the page is scrolled to.

## Help Requested

- Any constructive feedback is more than welcome!

## Checklist


will0684 (changes requested):
Awesome work 👍 

If you check out Paul's issue here: https://github.com/DTS-STN/Alpha-Site/issues/140

He brings up that we should include **focusing** the element that is scrolled into view as well. Best practice and better a11y that way.

nibivi77:
> Awesome work 
> 
> If you check out Paul's issue here: #140
> 
> He brings up that we should include **focusing** the element that is scrolled into view as well. Best practice and better a11y that way.

nibivi77:
> Awesome work 
> 
> If you check out Paul's issue here: #140
> 
> He brings up that we should include **focusing** the element that is scrolled into view as well. Best practice and better a11y that way.

Addressed in the previous commits.

var-kyle (changes requested):
When clicking on an item in the error box, can you make it so that the focus is put on the input of the linked error? Currently the focus remains on the the list of errors so pressing Tab scrolls the page back to the top.

nibivi77:
> When clicking on an item in the error box, can you make it so that the focus is put on the input of the linked error? Currently the focus remains on the the list of errors so pressing Tab scrolls the page back to the top.

Good catch, I will work on this!

PR #160, opened 2021-07-06
Disability question is required on yes.

nibivi77:
# Description

[Required disability](https://trello.com/c/Ekn7HPRY/328-for-the-disability-question-on-the-sign-up-page-change-logic-and-text-for-the-follow-up-question-asking-about-what-accommodation)

If the user selects "yes" for the disability question in the signup page, then the text field is required.

## Acceptance Criteria

Must enter some text in the yes text box.

## Test Instructions

1. Navigate to branch name: `required-disability`
2. Go to the sign up page
3. Submit the form as is. (all empty)
4. Should see 4 errors.
5. Now, select yes for the question: "Do you identify as a person with a disability?"
6. Re-submit the form.
7. Should see 5 errors.

## Help Requested

- any constructed feedback is greatly appreciated!

## Checklist

### DIFF
--- a/components/molecules/OptionalTextField.js
+++ b/components/molecules/OptionalTextField.js
@@ -70,6 +70,7 @@ export function OptionalTextField(props) {
             }
             dataTestId={props.textFieldDataTestId}
             dataCy={props.textFieldDataCy}
+            error={props.error}
           />
         ) : (
           <TextField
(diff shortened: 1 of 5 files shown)


EphraimAssi (changes requested):
Looks good, the only small problem is that the question after answering yes still says optional.
image
That could lead to confusion.

nibivi77:
> Looks good, the only small problem is that the question after answering yes still says optional.
> image
> That could lead to confusion.

Thank you I added the required notification as needed. The question itself is optional, but if you choose to answer with a "yes", then it becomes required.

EphraimAssi (approved):
Sorry, I was on the wrong branch, but looks good!

PR #162, opened 2021-07-06
Moved error box close to form.

nibivi77:
# Description

There is no task for this one, it was found in the a11y audit.

## Acceptance Criteria

- Error-box should be right on top of the form.

## Help Requested

N/A

## Checklist


var-kyle (changes requested):
I have to admit that it's not really clear where to place the error box, but I do think that the link to the privacy policy should be kept above the error box so that it's closer to the text talking about what we collect.

image

var-kyle (approved):
Looks great! Thanks @nibivi77!

PR #163, opened 2021-07-06
Accessibility audit - home page lacks meaningful heading structure

var-kyle:
# Description

From the accessibility audit:

Page lacks meaningful heading structure | Home | On this page, there are two heading level 1 "Service Canada Labs" and "Explore new services and help improve them". Both are acting as the main heading of the page.  Identify the main heading and update the second heading level to `<h2>`.  Also, the "Skip to main content" link should point to that heading which should be placed in the "Main" region of the page.
-- | -- | --

This PR moves the Banner component into the `<main>` region of the Layout component and updates to previous `<h1>` on the homepage to an `<h2>`. The skip to main content button links to the heading in the banner.

## Acceptance Criteria

Page structure updated as recommended from the accessibility audit

## Test Instructions

1. Navigate to the home page
2. Verify that there is only one `<h1>` and that the skip button links to it

## Checklist

## Product and Sprint Backlog

### DIFF
--- a/components/atoms/Banner.js
+++ b/components/atoms/Banner.js
@@ -10,10 +10,7 @@ export const Banner = ({ siteTitle, headline }) => {
     <div title="Home banner" className="bg-banner-img py-8">
       <div className="lg:container xxs:mx-0 xxs:px-0 lg:mx-auto lg:px-6 xxl:mx-auto">
         <div className="xxs:w-screen lg:w-2/3 xl:w-1/2 bg-dk-blue bg-opacity-90 text-white p-4">
-          <h1
-            className="text-h1 font-medium pt-4 pb-2 break-words"
-            id="pageMainTitle"
-          >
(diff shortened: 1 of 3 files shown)


nibivi77 (approved):
I feel like it would be better to skip the banner when "skipping content". Regardless, this should fix the issue in the audit.

Thank you!

will0684 (changes requested):
I don't think we should be putting the banner in the main block. We have plans to bring in the DECD Design System later and it's Layout component is based on what we already had. It just doesn't make sense to me to skip to the title of the page in the banner when a screen reader user already knows the title of the page from the page title itself. I think we should just keep the banner heading an H1 and make the content heading an H2 and skip to the content heading.

var-kyle:
Good call @will0684, I misunderstood and thought the skip link had to point to the `<h1>` but that wasn't the case at all.

will0684 (approved):

PR #170, opened 2021-07-08
Validation error redirect

nibivi77:
# Description

- Invalid CUID in the URL when validating the email sends the user to the error page with a "Wrong URL" message. (French is on the same page)
- Expired CUID (not found in the DB) when validating the email sends the user to the error page with a "Expired URL" message. (French is on the same page)

For temporary emails, use this link: https://temp-mail.org/en/

## Acceptance Criteria

- When passing in a CUID of less than 25 characters, show "Wrong URL".
- When passing in an expired CUID , show "Expired URL".
All strings on the error page are hard-coded. Reason:
- Both French and English are shown on the same page at the same time.

## Test Instructions

1. Sign up with an email.
2. Go in the email's inbox and copy paste the email in the URL
3. Remove a character from the URL to make it shorter.
4. error page "Wrong URL" should show up.
5. To test the "Expired URL", you need to delete the CUID in the database BEFORE clicking the validation link in the email's inbox.
6. Make sure to test all browser widths. (Mobile as well)

## Help Requested

- What if the CUID is 26 characters or more?
- Can I create unit or integration tests for this? If so, where? ex.: When sending a short CUID, the URL should contain `/error`

## Checklist

## Product and Sprint Backlog

[Wrong URL and Expired URL error pages](https://trello.com/c/JCL7rdNg/339-wrong-url-and-expired-url-error-pages)


var-kyle (changes requested):
When I first wrote the page, I didn't realize you could read the querystring and use `getStaticProps`, but you can by using `next/router`.  Could you change it to a statically generated page and move where we read the querystring into the page?

var-kyle (dismissed):
I hadn't considered that unit tests would start failing if we changed this to a statically generated page, so we can keep it as it is for now. Nice work on the error messages though.

will0684 (changes requested):
I think for if the cuid is more than 25, also show wrong link. Also can we remove commented out code if it's no longer going to be used? Nice work 👍

will0684 (approved):
Tested in the sign up flow and its working well, nice work! 👍

PR #174, opened 2021-07-09
Moved scroll property at the right location.

nibivi77:
# Description

When the form displays error like: "email already exists" or "Unknown error", the page would not scroll up on its own.

This PR ensures the page scrolls up after any error.

## Test Instructions
There are 2 ways to test this.
1. The first, is to `npm run dev`. This means that your first signup will throw this known bug:
 image

2. Sign up with an already existing email to get this:
 image

## Checklist


var-kyle (approved):
Awesome! It's to have this fixed, it was kind of annoying not having the scroll on certain errors.

PR #175, opened 2021-07-09
Add TTL to new mongo records

var-kyle:
# Description

[Add TTL to new Mongo records](https://trello.com/c/fp0IfI41)

When a user signs up, add an expiry date to the new record. When that date is reached, the record will be automatically deleted. When a user verifies their email, the expiry date is removed, so the record won't be automatically removed.

## Acceptance Criteria

Records are automatically deleted after 24 hours without validating their emails. When the email is validated, the corresponding record shouldn't be deleted.

## Test Instructions

1. Go through the signup flow without validating
2. Wait 24 hours
3. Try the link after waiting and see that you get an expired link error

## Checklist

## Product and Sprint Backlog

### DIFF
--- a/lib/mongodb/connect.js
+++ b/lib/mongodb/connect.js
@@ -47,6 +47,11 @@ export async function connectToDatabase(mongoURI, mongoDB) {
       unique: true,
       partialFilterExpression: { email: { $exists: true } },
     },
+    {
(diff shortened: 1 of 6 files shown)


will0684 (approved):
Nice work! 👍

PR #186, opened 2021-07-13
All text fits in pages.

### LINKED ISSUE #180: Bad or no reflow on some screen resolutions
On the French page, the page title "À propos des laboratoires" doesnt wrap to fit the page. The only visible part is "À propos des labo".

To see the issue, view the page using Firefox responsive design mode, screen size 320.

Ensure that all content falls in a single column on smaller screen sizes.

nibivi77:
# Description

fixes #180 

Fixed texts where it would expand beyond the page on small screens.
Also fixed an issue in the confirmation page where a div was inside a p tag.

## Acceptance Criteria

All texts are contained within the pages width.

## Test Instructions

1. Navigate throughout the app on small screens to see the text.

## Checklist

### DIFF
--- a/components/molecules/Experiment.js
+++ b/components/molecules/Experiment.js
@@ -13,7 +13,7 @@ export const Experiment = (props) => {
   };
   return (
     <a
-      className={`shadow-experiment-shadow p-4 border-b-4 min-h-400px ${
+      className={`shadow-experiment-shadow p-4 border-b-4 h-400px ${
         "border-" + (tagColours[props.tag] || "gray-experiment")
       }`}
       data-testid={props.dataTestId}
--- a/components/molecules/Experiment.js
+++ b/components/molecules/Experiment.js
@@ -13,7 +13,7 @@ export const Experiment = (props) => {
   };
   return (
     <a
-      className={`shadow-experiment-shadow p-4 border-b-4 h-400px ${
+      className={`shadow-experiment-shadow p-4 border-b-4 min-h-400px ${
         "border-" + (tagColours[props.tag] || "gray-experiment")
       }`}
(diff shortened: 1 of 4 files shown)


EphraimAssi (dismissed):
Looks good!

components/molecules/Experiment.js
```diff
@@ -13,7 +13,7 @@ export const Experiment = (props) => {
   };
   return (
     <a
-      className={`shadow-experiment-shadow p-4 border-b-4 h-400px ${
+      className={`shadow-experiment-shadow p-4 border-b-4 min-h-400px ${
```
will0684, line 16: Is this because text is falling outside the card @nibivi77 ?

will0684 (changes requested):

components/molecules/Experiment.js
will0684: I want to clarify because Xiaopu wants text that doesn't fit in the card to be cut off and replaced with a "...". Similar to the CDS blog here when a description exceeds the size of the card:

https://digital.canada.ca/blog/

nibivi77: Right now, we have this on 320px:
image

With that `min-h-400px` it gives this:
image

nibivi77: An alternative is to have the box stay 400px and have it scroll-able if the text is longer than the box.

nibivi77: The issue with the `overflow-ellipsis` is that it will cut out some info. and if we want to expand it, it will enlarge the card regardless.

will0684 (approved):
Nice work thanks! 👍

PR #195, opened 2021-07-14
Add a .dockerignore file

pcraig3:
We want to prevent unnecessary config or documentation files that we don't anticipate from being added to the container.

# Description

Added a `.dockerignore` file so that we don't end up adding `.svn` or `.git` folders to the container unintentionally. I tested this locally by building the container, but we can't really know if it works until it's deployed to prod. But we can know that it doesn't break anything.

#inspo for the `.dockerignore` file is here: https://github.com/pcraig3/hols/blob/main/.dockerignore

Trello card for this is here: https://trello.com/c/gBRWOzvN/390-390-add-a-dockerignore-file-to-the-repo-so-that-we-dont-unintentionally-build-containers-with-git-svn-folders-in-them

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,23 @@
+# git stuff
+.git
+.gitignore
+.github
+
+# svn as well
+.svn
+
+# project stuff
+node_modules/
+npm-debug.log
+.DS_Store
+*.env
+coverage/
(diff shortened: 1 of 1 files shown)


will0684 (approved):
Looks good thanks!

PR #196, opened 2021-07-14
QA about page

nibivi77:
# Description

[#371 About Page EN/FR](https://trello.com/c/vxcJ4mB2/371-371-about-page-en-fr)

Converted our old design to the new one in Figma under "QA list pre-launch".

## Acceptance Criteria

The about page should reflect the Figma design.

## Test Instructions

1. Navigate to the about page and test different screen sizes both in French and English.

## Checklist


components/molecules/CallToAction.js
```diff
@@ -13,7 +13,10 @@ export function CallToAction(props) {
       <div className="layout-container pb-10 pt-10 text-xs md:text-base">
         <h2>{props.title}</h2>
         <div className="flex flex-col lg:grid lg:grid-cols-2 lg:gap-24 gap-5">
-          <p>{props.description}</p>
+          <p>
+            {props.descriptionP1} <br />
+            <br /> {props.descriptionP2}
```
will0684, line 18: Part of Government of Canada design guidelines is to never use <br> tags

nibivi77: Humm ok sounds good. ill just put some mb-4

components/molecules/CallToAction.js
```diff
@@ -46,9 +50,14 @@ CallToAction.propTypes = {
   title: PropTypes.string.isRequired,
 
   /**
-   * a short description about what the call to action is about
+   * a short description about what the call to action is about - part 1
+   */
+  descriptionP1: PropTypes.string.isRequired,
```
will0684, line 55: To keep this component simple, I would recommend implementing the description this way:

1. In CallToAction.js, remove the `<div>` on line 16 that is wrapping the `<p>` (This div is not useful)
2. Add this to the `<p>` tag:
```html
<p style={{ whiteSpace: "pre-line" }}>{props.descriptionP1}</p>
```
3. Combine the two strings in the translation files and separate them with two newline characters `\n` in that single string like so: 

"You’re invited to test our experimental ideas and participate in research interviews to make Service Canada better for everyone. Every bit of feedback helps us make sure we’re making our services simple and easy to use.\n\nYour participation will not affect your access to government services. You can unsubscribe at any time."

will0684 (changes requested):

components/molecules/CallToAction.js
will0684: This way we can have 1 description prop and put whatever string length or paragraphs you want.

components/molecules/CallToAction.js
will0684: My comment on how to use the single description prop for multi-line text should address this as well.

components/molecules/CallToAction.js
nibivi77: Ya that makes much more sense. I was thinking about that but I did not know if we could do \n in the lang files

EphraimAssi (commented):
Looks good overall, just some really minor details.
image
I'm not sure if you can see it clearly in this picture, but we have the text "if you have..." in the gray area of the list, while it's under that in the Figma design. 
Figma design: 
image

Also, we have an unsubscribe link under the sign-up button, but on Figma it's directly in the text.
Figma design: 
image

What we have:
image

components/molecules/CallToAction.js
```diff
@@ -8,12 +8,27 @@ import { useTranslation } from "next-i18next";
  */
 export function CallToAction(props) {
   const { t } = useTranslation("common");
+  const texts = props.description
+    .split(/(unsubscribe|désinscrire)/)
+    .filter(Boolean);
```
will0684, line 13: I am not sure why you are using a Boolean constructor as the filter method but I think we should avoid configuring this component in a way that it's behaviour is dependent on a specific value in the string being passed in. The component should not at all be concerned with any specifics of the data. I think it's a better idea to use the html or markdown string method as it removes this logic entirely and makes it so we can pass any string in and see a predictable output.

will0684 (changes requested):

pages/about.js
```diff
@@ -75,9 +66,29 @@ export default function About(props) {
...
+          .
+        </p>
+      </section>
       <CallToAction
         title={t("signupTitle")}
-        description={t("becomeAParticipantDescription")}
+        description={
+          <>
+            {t("becomeAParticipantDescriptionStart")}
+            <Link href="/unsubscribe">
```
will0684, line 86: So because now your passing in raw JSX and HTML we are getting an error with the prop type being of type "object" when it is expecting type "string". This does not mean changing the prop type to an object because then we could pass in literally anything and we don't want that. It should stay a string and work in the way that the TextButtonField works in that you can pass in either HTML (as a string) or any other string.

will0684 (changes requested):

pages/about.js
nibivi77: I read about that and it opens vulnerability to cross-site scripting forgery attacks, that's why its called `dangerouslySetInnerHTML`. Also passing in raw HTML disables us from passing JSX an/or components.

So this won't work:
`description="{t(`becomeAParticipantDescriptionStart`)}<Link href='/unsubscribe'><a className='underline'>{t('unsubscribeLinkText')}</a></Link>{t('becomeAParticipantDescriptionEnd')}"`

results in this:
image

will0684: We don't have to worry about cross site scripting because our content is pre-built and pre-rendered on the server before it gets to the client and there is no way for our traffic to be intercepted. The client side will never have access to that, they will only ever see the completed HTML as they are being served a static page. Rogue javascript from other sources has literally no way of getting into that element.

will0684: Essentially an attacker would have to have access to our server to inject their own JS into it. Which they never will.

will0684: And to your second point, you write out the HTML as a string, like in the example of the html prop in the TextButtonField storybook. The string would be in the translation files.

will0684 (approved):
Nice work Benoit! 🚀

PR #197, opened 2021-07-15
Update mongo indexes

var-kyle:
# Description

CosmosDB does not support partial or sparse indexes, which we were using to keep a unique constraint on emails, and allow us to delete them. Without a partial or sparse index, two records with removed emails would be considered the same and fail the unique constraint. To get around this issue and keep emails unique, when a user unsubscribes, I am setting the email field to the cuid, which will be unique.

## Acceptance Criteria

When a user unsubscribes, there are no unique constraint errors.

## Test Instructions

1. Follow the signup flow with 2+ emails
2. Unsubscribe for 1 of them
3. Unsubscribe for a second and notice that no errors appear

## Checklist

## Product and Sprint Backlog

### DIFF
--- a/lib/mongodb/connect.js
+++ b/lib/mongodb/connect.js
@@ -47,11 +47,10 @@ export async function connectToDatabase(mongoURI, mongoDB) {
       key: { email: 1 },
       name: "email",
       unique: true,
-      sparse: true,
     },
     {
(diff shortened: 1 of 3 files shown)


will0684 (approved):
Looks good!

PR #199, opened 2021-07-15
Adding Phasebanner back.

nibivi77:
# Description

Following the QA list pre-launch and adding the Phasebanner back on all pages.
image

## Acceptance Criteria

Phasebanner must be on all pages except the splash page.

## Test Instructions

1. Load any page and test the Phasebanner at the top of the page on any width and both languages.

## Checklist

### DIFF
--- a/components/atoms/PhaseBanner.js
+++ b/components/atoms/PhaseBanner.js
@@ -12,7 +12,7 @@ export const PhaseBanner = ({ phase, children }) => {
         <span className="font-body text-xs text-circle-color font-extrabold flex-shrink-0 bg-white px-4 py-1 my-auto leading-6">
           {phase}
         </span>
-        <p className="font-body text-xs mt-5 sm:mt-auto text-white sm:ml-4 pt-1 my-auto">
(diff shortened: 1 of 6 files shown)


components/atoms/PhaseBanner.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+export const PhaseBanner = ({ phase, children }) => {
+  return (
+    <div className="bg-circle-color">
+      <div className="block sm:flex py-4 layout-container">
+        <span className="font-body text-xs text-circle-color font-extrabold flex-shrink-0 bg-white px-4 py-1 my-auto leading-6">
+          {phase}
+        </span>
+        <p className="font-body text-xs mt-5 sm:mt-auto text-white sm:ml-4 pt-1 my-auto">
```
will0684, line 15: Looking real sharp. All I would recommend is that we vertically center the text within the space that it has so it lines up right along the horizontal center axis of the "Test site" block.
image

nibivi77: `align-middle` did not work so I added bottom padding of 1px at large screens. it does the trick. :)

var-kyle (approved):
Looks great, nice work!

PR #200, opened 2021-07-16
Fix TTL index in CosmosDB

var-kyle:
# Description

Simplified the TTL index for Cosmosdb and set the correct max value. Had to add a check for when doing unit tests because, while the int32.max works in the real MongoDB, it makes the unit tests fail.

## Acceptance Criteria

The TTL index is actually created in CosmosDB.

## Checklist

## Product and Sprint Backlog


nibivi77 (approved):
:+1: Très bien :eyes:

PR #203, opened 2021-07-16
QA landing page Design

nibivi77:
# Description

[QA landing page Design](https://trello.com/c/iFKKej8H/370-370-landing-page-en-fr)

Updated home page with the new QA list pre-launch designs.

## Acceptance Criteria

The landing page must match the Figma design under "QA list pre-launch"

## Test Instructions

1. Go on the home page and swap languages with multiple screen width.

## Checklist

## Questions

1. Should I rename all the "experiments" to "projects" in this PR? or in a separate one.


nibivi77:
On the home page, there is a list with 3 elements which is displayed like this:
image

However, I could do something like this instead, thoughts?:

image

will0684:
> On the home page, there is a list with 3 elements which is displayed like this:
> image
> 
> However, I could do something like this instead, thoughts?:
> 
> image

I think it would be better to go with your second choice there.

nibivi77:
> > On the home page, there is a list with 3 elements which is displayed like this:
> > image
> > However, I could do something like this instead, thoughts?:
> > image
> 
> I think it would be better to go with your second choice there.

Done :)

will0684 (approved):
Looks great on desktop and mobile good work! 🚀

PR #206, opened 2021-07-16
QA signup page

nibivi77:
# Description

[Signup page](https://trello.com/c/sqFFyn83/373-373-sign-up-pages-en-fr)
Updated the sign up page to look like the design on Figma.
This PR covers mostly content and font sizing.

## Important note
- There is a field missing, I will create a separate PR for that field as the current PR is getting heafty.

## Acceptance Criteria

The Sign up page must look like the design on Figma QA list pre-launch.

## Test Instructions

1. Go on the sign up page and play with the screen size. 
2. Verify that both languages match the Figma design.

## Checklist


will0684 (changes requested):
Looks good! Only issue I see is a bit of the text at the bottom of the form covers other elements when the device is small. Here it is on an iPhone 5 dispay:
image

On an iPhone 10 you can see as well that it needs some spacing:
image

var-kyle (changes requested):
Would you be able to add just a bit more spacing between the consent question and the privacy link? On iPhone 5 it is very close:
image

Everything else looks great though!

will0684 (changes requested):
Nice work definitely a lot better, but I think we could use a small bit of more space beneath the consent and above the privacy policy link

image

will0684 (approved):

will0684:
Ignore my last comment, this should be fine.

var-kyle (approved):
Looks awesome, thanks @nibivi77!

PR #210, opened 2021-07-19
QA-unsubscribe

nibivi77:
# Description

[#377 Unsubscribe Pages EN/FR](https://trello.com/c/97qjL7SL/377-377-unsubscribe-pages-en-fr)

Updated the unsubscribe translations. French and English.

## Acceptance Criteria

The text must match the Figma design.

## Test Instructions

1. Start by subscribing a email
2. once subscribed, follow the unsubscribe flow in both french and english.
3. Make sure all text matches the Figma design as well as the sizing of the text.
4. All text must also fit within the page.

## Help Requested

- I need help modifying the content sent back to the user in the unsubscribe email.
**What we have:**
image

**What we need:**
image

## Checklist


var-kyle (changes requested):
There is a `resend email` link that shouldn't be there on the `/thankyou` page.

image

And I noticed that the title on the unsubscribe page is `Service Canada Labs - Sign-up` and it should be `Service Canada Labs - Unsubscribe`.

Everything else looks good! I can change the email contents for you.

pages/thankyou.js
```diff
@@ -79,7 +79,15 @@ export default function Confirmation(props) {
...
-            <ActionButton id="resend_email" text={t("resendEmail")} tertiary />
+            {referrer === "signup" ? (
+              <ActionButton
+                id="resend_email"
+                text={t("resendEmail")}
+                tertiary
+              />
+            ) : (
+              ""
+            )}
```
var-kyle, line 90: I think the resend button can just be completely removed, it doesn't look like it's used anywhere.

var-kyle (changes requested):
Great work @nibivi77! Just one little thing with the resend button.

pages/thankyou.js
nibivi77: Ah yes you are right. I thought we used it when subscribing. Removing...

var-kyle (approved):
Nice work, looks great!

pages/about.js
```diff
@@ -78,10 +78,10 @@ export default function About(props) {
         </p>
       </section>
       <CallToAction
-        title={t("signupTitle")}
+        title={t("signupTitleCallToAction")}
         html={t("becomeAParticipantDescription")}
         href={"/signup"}
-        hrefText={t("signupTitle").slice(0, t("signupTitle").length - 1)}
+        hrefText={t("signupBtn")}
```
will0684, line 84: I think we need this and text in the Experiments page CallToAction as well

will0684 (changes requested):

pages/about.js
nibivi77: You are right! good catch

pages/experiments.js
```diff
@@ -110,7 +110,7 @@ export default function Experiments(props) {
         title={t("signupTitle")}
```
will0684, line 110: This should be signupTitleCallToAction too I believe?

will0684 (approved):

PR #221, opened 2021-07-21
Changed path of VC to virtual-assistant

DaceyTom2:
# Description

Path for /projects/virtual-concierge changed to /projects/virtual-assistant


will0684 (approved):
👍

PR #242, opened 2021-07-22
Update some of the DC images in french

pcraig3:
# Description

We want the Digital Center images to be a uniform size, but some of the French images were fullscreen rather than 16:9 or whatever the ratio is.

This PR swaps some of the French images for cropped versions, but makes no functional changes.

## Screenshots

| before | after |
|--------|-------|
|  some of the images are really long      |  all the images are the same size      |
|   Screen Shot 2021-07-22 at 13 47 40  |  Screen Shot 2021-07-22 at 13 47 13  |


var-kyle (approved):
Looks much better!

PR #245, opened 2021-07-22
Scenario tag changed to comply with accessibility

DaceyTom2:
# Description

h2 tag for Scenario changed to a p tag as it is more suited as content than a heading. This was done to comply with accessibility requirements.

## Test Instructions

1. Navigate to virtual-assistant page
2. "Scenario" should now be in a p tag with same styling


pcraig3 (approved):
Looks the same as before 🔍

PR #255, opened 2021-07-23
Create consistent page titles, remove unused keys from common.js

pcraig3:
# Description

Okay, so there's ~3~ 4 things going on in this PR:

### 1. Consistent page titles

As a rule, you want to use the same page title in the `<title>` and the `<h1>` element. The only difference is that in the `<title>` you also want to include the site name so that Screen readers have context. As with most things related to screen readers, you want to put the page title in front of the name of the site, so that the more important information is read before the generic information. We are also sending page titles to Adobe Analytics as well, so I've used the same as what the `<title>` is to send to AA.

To take the "About these labs" page as an example:

- `<title>`: `About these labs — Service Canada Labs`
- Header meta (“dcterms.title”) for Adobe Analytics: `About these labs — Service Canada Labs`
- `<h1>`: `About these labs`

Since we want the "h1" content to match the "title", I've also removed a bunch of our specific "scLabsAboutTitle" keys because we don't use them anymore.

### 2. Remove most of the no-longer-used keys in `common.js`

We had a bunch of keys in there that we don't use anymore, so I ctrl+f'd for the keys and removed almost all the ones that aren't referenced anywhere in the app.

### 3. Duplicated alt text on the Digital Channel Overview page

I accidentally duplicated the text here when I was putting it into that page. Easy fix.

### 4. Use `’` instead of `'` for apostrophes.

You see them a lot in French, because when there are vowels, you drop the vowel in the article (eg "j’ai faim"). But also you have them a lot in English for contractions. 

- Good: `J’ai faim`, `I can’t even`
- Bad: `J'ai faim`, `I can't even`

Basically, the idea is that for web typography, you should use curly quotes (or ”[smart quotes](https://smartquotesforsmartpeople.com/)”).


will0684 (approved):
Looks good! Thanks Paul!

will0684 (approved):
Cheers Paul 🚀

PR #259, opened 2021-07-26
Move "Report a Problem" bullets

pcraig3:
## Description

On our first go-around with the a11y group, they recommended we move the bullets in the Report a Problem widget _above_ the checkboxes, so that users could see what they were agreeing to before submitting. That was a "suggestion", not a "fail".

Then we moved them above the checkboxes, and they raised **2 issues** about them, and both are "fails".

1. the list items are not part of the form, so they should be outside the form.
2. a screen reader in “form” mode, would miss them

“Don’t have them in the form at all”, seems pretty illogical to me. It doesn't make sense to put them after the submit button or before the legend. So I'm moving them back to where they were before, when they weren't failing.

### Screenshots

| before | after |
|--------|-------|
| The bullets are above the checkboxes       |  The bullets are below the checkboxes     |
|   Screen Shot 2021-07-26 at 13 19 17   |    Screen Shot 2021-07-26 at 13 18 42   |

### DIFF
--- a/components/organisms/Layout.js
+++ b/components/organisms/Layout.js
@@ -1,5 +1,4 @@
 import PropTypes from "prop-types";
-// import Head from "next/head";
 import { Banner } from "../atoms/Banner";
 import { Menu } from "../molecules/Menu";
 import { Footer } from "./Footer";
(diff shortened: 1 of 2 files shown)


EphraimAssi (approved):
Looks good!

Issue #230, opened 2021-07-21
Enhancement: More appropriate semenatic markup could be added to How we work section


neilmispelaar:
### Description

The "How we work" section on the "about" page uses an ol (ordered list) to mark up the list. Immediately after the list item, a number is written in. (This is better described in the screenshot below where you can see two number ones, two number two's, two number three's, etc).

One issue is the "stutter effect - screen readers will read out the numbers twice: "List item with four elements. Element number one. One. Displaying our early works in progress. List item number two. Two. Encouraging people of diverse backgrounds....". Not the end of the world but not a perfect experience for screen readers.

The second issue is that ol (ordered list) are reserved for lists of sequence. These are generally used for recipes or instructions, where you have to do step 1, before going on to step 2, and so on and so forth. How the page reads (to me anyway), is that this is more of a general list of things that we do, and we don't do these things in sequence, or rather the sequence of events doesn't matter. We may blog before we use new technology.

### How to fix

Changing the list to an ul (unordered list where the sequence doesn't have meaning) and using an aria role=presentation on just the numbers in the list item could enhance the semantic markup.

### WCAG Compliance 

https://www.w3.org/WAI/WCAG21/Techniques/html/H48

### Screenshot

<img width="1700" alt="Screen Shot 2021-07-21 at 1 12 39 PM" src="https://user-images.githubusercontent.com/9325038/126535264-6683655c-04fc-48f9-a35b-66d0029f5f17.png">

PR #262, opened 2021-07-26
Accessibility on "How we work" section on the About page

shawn320:
# Description
The purpose of this pull request is to fix issue #230. 

- Change the list from ol to ul
- Add aria role of presentation on the numbers in the list 

## Checklist


components/molecules/List.js
```diff
@@ -6,7 +6,7 @@ import PropTypes from "prop-types";
 export function List(props) {
   let opacity = 40;
   return (
-    <ol className={props.className}>
+    <ul className={props.className} role="presentation">
```
var-kyle, line 9: My interpretation of the fix is that the `role="presentation"` should just be on the number, so rather than adding it in the `<ul>` it would be put in the `<span>` in each `<li>`

var-kyle (changes requested):
Looks good, but I think the `role="presentation"` needs to be moved.

components/molecules/List.js
shawn320: Thank you Kyle, I have moved the role to ```<span>```

var-kyle (dismissed):
Awesome, thank you @shawn320!

will0684 (approved):
Nice work Shawn, thanks!

PR #263, opened 2021-07-27
Updates for Digital Center overview page, Call to action, and "Projects" page

pcraig3:
# Description

There's 4 things happening in this PR.

### 1. Updating the content for the Digital Channel page

This pull request updates the English and French content on the Digital Channel project page, as well as adds some extra link styling.

### 2. Updates the CtA title and button text on the pages where it was out of date

We don't want the "step 1 of 3" on the bottom of our pages.

### 3. Updates the Projects page

- Make sure the phase tags are getting translated (previously, they weren't)
- Update the CSS for the "experiments" boxes since they had a max-height and then the text would leave the box
- Add some more bottom margin under breadcrumbs so that the h1s have a bit more space

### 4. Adding non-breaking spaces in front of colons for French language text

In French, you want to have a non-breaking space between colons and the words in front of them, so that they don't end up orphaned on the line below.

### Screenshots

| before | after |
|--------|-------|
| text leaves the box when it expands in French | text stays in the box, in both languages |
|  Screen Shot 2021-07-27 at 10 27 04  | Screen Shot 2021-07-27 at 10 46 14   |

#### Translated the filters
| before | after |
|--------|-------|
| "All" isn't translating in French | "All" becomes "Tout" |
|   Screen Shot 2021-07-27 at 10 41 21    |   Screen Shot 2021-07-27 at 10 59 22    |

| before | after |
|--------|-------|
|   h1 is very close to the breadcrumbs     |  h1 is sensibly spaced out     |
|   Screen Shot 2021-07-27 at 10 46 38   |   Screen Shot 2021-07-27 at 10 46 29    |


will0684 (approved):
Headings look much nicer :) thanks!

PR #264, opened 2021-07-27
Feature/email csv

var-kyle:
# Description

[Email CSV](https://trello.com/c/XE5jvIGL)

We want to generate a .csv and send it in an email daily. To do this, I've set up an api route `/api/generate-csv` that will:
* get the data we want, 
* convert it to a .csv, 
* attach it to an email and send it

To make sure that no one can just hit that api endpoint to generate the .csv, I am checking that they have the correct authorization header in the request. The request is ignored if they don't have or have an incorrect authorization header.
image

The code on it's own will not send the email, but once this is merged in we can setup a pipeline in Azure DevOps that will trigger once a day that will hit this endpoint and send the email.

We are using `NOTIFY_REPORT_A_PROBLEM_EMAIL` as the destination email address and I realize there may be some confusion about that, but the email is correct and changing the env is kind of a hassle at this point so close to launch.

### New ENVs
```
USER_CSV_TEMPLATE_ID=
CSV_AUTH_KEY=
```

## Acceptance Criteria

A .csv is sent in an email when an authorized request is made to the new api endpoint.

## Test Instructions

1. Add some value to `CSV_AUTH_KEY` in your local .env
2. Add the `USER_CSV_TEMPLATE_ID` in your local .env
3. Using Postman, set the url to `http://localhost:3000/api/generate-csv`
4. Set the method to `POST`
5. Under Headers, set `Content-Type` to `application/json`
6. Under Headers, set `Authorization` to the value you entered for `CSV_AUTH_KEY`
7. Click send
8. Assuming you've set `NOTIFY_REPORT_A_PROBLEM_EMAIL` to an email address you have access to, you should see a new email show up with an attached .csv

## Checklist

## Product and Sprint Backlog


will0684 (approved):
Tested all the way through and its working great 🕺

PR #239, opened 2021-07-22
Add lang tags to language links, and update splash page lang selector for horizontal screens 

pcraig3:
# Description

There's 3 things going on in this PR:

1. Adding lang link to the language link at the top of the page. This is based on feedback we've gotten from the accessibility audit.
2. Update the "alt" text for the wordmarks on the splash page. On the splash page, we don't know if someone is going to select English or French yet, so we use both.
3. Update the splash page language selector so that it is always visible on horizontal screen sizes. Resolves #191 

Here's a little gif of the splash page

Screen Recording 2021-07-22 at 11 48 19

## Acceptance Criteria

Run the app, check that the lang tags work and the landscape splash page looks good.

PR #268, opened 2021-07-27
Remove bg blur on splash page

pcraig3:
# Description

My bad, I copied some CSS for #239 and it included blurring the background image. Removing the background blur on the splash page.

### Screenshot

| before | after |
|--------|-------|
|    Screen Shot 2021-07-27 at 11 21 22    |  Screen Shot 2021-07-27 at 11 21 02  |

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -11,7 +11,7 @@ export default function Index(props) {
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


will0684 (approved):
Looking good 👍

PR #273, opened 2021-07-27
Analytics page load

EphraimAssi:
# Description

[#380 Make sure page loads are being tracked after the first page load for analytics](https://trello.com/c/Z8tmWpF0/380-380-make-sure-page-loads-are-being-tracked-after-the-first-page-load-for-analytics)

I manually added a page load event on each to activate analytics when we switch pages, we can't really test until it's merged in.

## Acceptance Criteria

On each new page load, adobeDataLayer should push "event: pageLoad" in a array.

## Test Instructions

1. Add adobe analytics to env file
2. Run the code and open up the console (some errors will show up because of the analytics running locally)
3. Check if each new page load pushes a page load (just write adobeDataLayer in the console)

## Checklist

## Product and Sprint Backlog


pages/404.js
```diff
@@ -4,10 +4,18 @@ import { serverSideTranslations } from "next-i18next/serverSideTranslations";
...
 import { ReportAProblem } from "../components/organisms/ReportAProblem";
 import { ActionButton } from "../components/atoms/ActionButton";
+import { useEffect } from "react";
 
 export default function error404(props) {
   const { t } = useTranslation("common");
 
+  process.env.NEXT_PUBLIC_ADOBE_ANALYTICS_URL
+    ? useEffect(() => {
+        window.adobeDataLayer = window.adobeDataLayer || [];
```
will0684, line 14: So I was curious about this and decided to do some reading and I think we should avoid useEffects conditionally but rather put the condition _inside_ the useEffect. Some details here:

https://reactjs.org/docs/hooks-rules.html

will0684 (changes requested):

pages/404.js
will0684: For all I know, it could be the reason that it wasn't working in the top-level app.js

will0684 (approved):
Good stuff nice work!

PR #276, opened 2021-07-27
bg-cover on banner for all screen sizes

will0684:
# Description

Request from design to use background-cover on screen sizes smaller than 450px. Previously, background-cover was only applied after the 450px breakpoint.

## Product and Sprint Backlog


EphraimAssi (approved):
Looks good!

PR #278, opened 2021-07-27
Updated translations and tests to support new tags

will0684:
# Description

New tags were decided on by PO and team for projects. They are as follows:

- Upcoming projects
- Current projects
- Past projects

The CMS has already been updated to support the new tags, and Projects already created in the CMS now have their tags changed to the appropriate new tag.

## Product and Sprint Backlog


pcraig3 (approved):
Tested it and it works 👍

Issue #229, opened 2021-07-21
Navigation links showing active link via presentation but has no corresponding markup for screen readers


neilmispelaar:
### Description

The Navigation links (in this example Sign Up) is underlined (conveying meaningful information via presentation) showing it is the current page the user is on, but in code there is no hidden text telling the user the same information (that this link is the link that the user is currently visiting).

### How to Fix
Simply need to add some hidden text to the active link such as:

```
<a><span class="sr-only">You are currently on this page: </span>Sign Up</a> 
```

### WCAG Compliance

This constitutes a fail of WCAG 1.3.1 
https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html

### Screenshot

<img width="1590" alt="Screen Shot 2021-07-21 at 1 39 10 PM" src="https://user-images.githubusercontent.com/9325038/126534559-24897e15-0fa7-419c-aad8-2a85ed6441fc.png">

pcraig3:
I think it's easier then adding that hidden text. I had a check what [Bootstrap does with its navigation menu](https://getbootstrap.com/docs/5.0/examples/navbars/#), and they add `aria-current="page"` to the currently-active link. 

So, we could make sure that's on the links without adding a sentence of hidden text.

neilmispelaar:
You definitely could do it with aria labels, the risk is that any assistive technologies that don't support ARIA roles won't announce anything. 

But from a compliance perspective, either technique nets a pass, it's just real world coverage would be increased using the hidden text technique.

Some notes on the aria-current values: 

- If on the current page: `aria-current="true"`
- If in the current set of pages (but not on the current page): `aria-current="page"`

#### Additional notes: 
https://www.w3.org/WAI/tutorials/menus/structure/#indicate-the-current-item
https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaCurrent
https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/current/#possible-values

pcraig3:
Thanks for the links, @neilmispelaar! It looks like it would be best to preceding text, as you mentioned, but I don't want to create a dodgy French translation, so I'm shipping the "aria-current" change for now.

PR #279, opened 2021-07-27
Menu links are underlined for child pages

pcraig3:
We still want the "Explore our projects" link underlined when someone is looking at a project.

From Xiaopu:
> Could you make sure that in the top menu, “Explore our projects” is underlined to indicate to user that they are in the projects section of the site?

Also, added an `aria-current="page"` attribute based on what Bootstrap does for its active nav items. Because it only seems to indicate the current page, not the current section, I am only including it if the URL is an exact match. It seems like it would be better to provide screen reader descriptions ("Current page: ", "Current section: "), but we can do this at a later time.

### DIFF
--- a/components/molecules/Menu.js
+++ b/components/molecules/Menu.js
@@ -43,16 +43,18 @@ export function Menu(props) {
 
       <ul id="menuDropdown" className="menuDropdown" role="menu">
         {props.items.map((item, key) => {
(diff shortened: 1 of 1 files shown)


pcraig3:
This partially addresses, but doesn't fully resolve #229

will0684 (approved):
Looks good! Thanks

PR #280, opened 2021-07-27
Standardize disclaimer across project pages

pcraig3:
Added a "this is a test site" disclaimer across the bottom of the Digital Channel and in the process, removed the separate lines from the other two locales files.

We want them all to say the same thing, so I coped the Life Journeys' disclaimer into the common locales file and deleted the local versions.


will0684 (approved):
Tested and looking good 👍

PR #281, opened 2021-07-28
Update get users function

var-kyle:
# Description

The `getUsers` function originally just got everything from the users collection in the database - this was wrong. This updates the function to only get verified users. I also added the projection option so that the query ignores the `_id` of each record which is unnecessary to include. Also added a unit test to verify that the correct users are added.

## Acceptance Criteria

Only verified users are put in the .csv.

## Checklist

## Product and Sprint Backlog


will0684 (approved):
Nice work!

PR #291, opened 2021-07-28
Using <strong> instead of CSS to bold

krischarbonneau:
# Description

I removed every use of `font-bold` (only one `p` element but all the headings had it, which isn't necessary since they're bold by default) and replaced with `<strong>` as per the below message from the re-audit:

```
Using changes in text presentation to convey information without using the appropriate markup or text

Bold text on the pages is styled using CSS to visually appear as bold and not using the appropriate HTML code.

Source code:
<p class="mb-6 font-bold">You cannot apply for services or benefits through this test site. Parts of this site may not work and will change.</p>
<p class="mb-6 font-bold">Vous ne pouvez pas demander de services ou de prestations par l’intermédiaire de ce site d’essai. Certaines parties du site pourraient ne pas fonctionner et seront modifiées.</p>
To fix this, use the HTML <strong> element to code bold text.
Please check all content.
```

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/projects/life-journeys`

## Checklist

### DIFF
--- a/pages/projects/life-journeys.js
+++ b/pages/projects/life-journeys.js
@@ -71,21 +71,17 @@ export default function LifeJourneys(props) {
         <section className="layout-container relative mb-10">
           <h1
             id="pageMainTitle"
-            className="mb-8 text-h1l font-bold flex-wrap"
+            className="mb-8 text-h1l flex-wrap"
             tabIndex="-1"
           >
             {t("lj:lifeJourneysTitle")}
           </h1>
-          <h2 className="mb-6 font-bold leading-10">
-            {t("lj:lifeJourneysHeading1")}
-          </h2>
+          <h2 className="mb-6 leading-10">{t("lj:lifeJourneysHeading1")}</h2>
           <p className="mb-6">{t("lj:lifeJourneysContent1")}</p>
           <p className="mb-6">{t("lj:lifeJourneysContent2")}</p>
           <p className="mb-6">{t("lj:lifeJourneysContent3")}</p>
           <p className="mb-6">{t("lj:lifeJourneysContent4")}</p>
-          <h2 className="mb-6 font-bold leading-10">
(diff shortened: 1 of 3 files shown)


pages/projects/life-journeys.js
```diff
@@ -71,21 +71,17 @@ export default function LifeJourneys(props) {
         <section className="layout-container relative mb-10">
           <h1
             id="pageMainTitle"
-            className="mb-8 text-h1l font-bold flex-wrap"
+            className="mb-8 text-h1l flex-wrap"
```
pcraig3, line 74: I was gonna say, we don't want to use "strong" in headings, but it looks like they are still bold, so all good.

pages/projects/life-journeys.js
```diff
@@ -124,11 +118,11 @@ export default function LifeJourneys(props) {
...
           <p className="mb-6">{t("lj:lifeJourneysContent11")}</p>
-          <h2 className="mb-6 font-bold leading-10">
-            {t("lj:lifeJourneysHeading4")}
-          </h2>
+          <h2 className="mb-6 leading-10">{t("lj:lifeJourneysHeading4")}</h2>
           <p className="mb-6">{t("lj:lifeJourneysContent12")}</p>
-          <p className="my-6 font-bold">{t("projectsDisclaimer")}</p>
+          <p className="my-6">
+            <strong>{t("projectsDisclaimer")}</strong>
+          </p>
```
pcraig3, line 125: Great! Let's do this on other pages as well.

pcraig3 (changes requested):
Thanks for this Kris! Could you do the same thing for the `digital-center.js` and `/virtual-assistant/index.js`? Both of them are doing the same thing with their disclaimer text (and DC has a bold h1).

pcraig3 (approved):
Let's ship this mf. 👍

PR #292, opened 2021-07-28
Fix CosmosDb ttl field

var-kyle:
# Description

The ttl field in CosmosDb is a duration, in seconds, but the way I wrote it before resulted in a time for the record to be removed:
image

Apparently the ttl in CosmosDb doesn't work this way, so this fixes that problem and records should be deleted after 24 hours.

## Checklist

## Product and Sprint Backlog


will0684 (approved):
Thanks 👍

PR #296, opened 2021-07-28
Menu links blue on desktop and mobile

pcraig3:
Super small PR. Just changing the links to blue in the top nav menu.


will0684 (approved):
Sorry for the confusion I may have caused on this. Thanks for the fix,

PR #297, opened 2021-07-29
Swap out images over 150kb for compressed images

pcraig3:
# Description

Some of our images were very large, so we want them to be a bit smaller.

This updates all the images for the Digital Centre (all are less than 100kb now), as well as the "Life journey map" image (which used to be 1.5 MB and is now ~120kb.

No functional changes. The site should look the same as previously.


will0684 (approved):
Thanks for this 👍

PR #299, opened 2021-07-29
Fix page title max width + image alt text

pcraig3:
# Description

This PR does 2 things:
 1. Removes `w-max` classes for all the headings, which fixes how page titles are displayed
 2. Changes an alt-text in French that was the wrong alt-text

## Removes `w-max` classes for all the headings

By setting a max-width, we are forcing the h1s to be a certain width, and then long page titles go off the screen. Removing the class entirely fixes that. 

| before | after |
|--------|-------|
|   Screen Shot 2021-07-29 at 10 59 03   |   Screen Shot 2021-07-29 at 10 59 05  |

### DIFF
--- a/pages/about.js
+++ b/pages/about.js
@@ -51,11 +51,7 @@ export default function About(props) {
           <meta name="dcterms.issued" content="2021-05-11" />
         </Head>
         <section className="layout-container relative mb-10">
-          <h1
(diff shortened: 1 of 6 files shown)


will0684 (approved):
This is great thank you

PR #301, opened 2021-07-29
Update alt text for French image 2.3 on Digital Centre page

pcraig3:
# Description

Very small change. It looks like I didn't update the alt text for this one image, and instead copy+pasted alt text from an earlier image.


EphraimAssi (approved):
Looks good!

PR #304, opened 2021-07-29
Sort by lable

shawn320:
# Description

Display minority list on the Signup page in alpabetical order for both English and French. 

- Added a sort function in ```OptionalListField``` component (credit to @var-kyle )
- fix spacing between options


components/molecules/OptionalListField.js
```diff
@@ -147,4 +156,9 @@ OptionalListField.propTypes = {
    * legend text for the list items
    */
   listLabel: PropTypes.string,
+
+  /**
+   * sort the children being passed in based on labels
+   */
+  sortedChildren: PropTypes.func,
```
var-kyle, line 163: You don't need to add this since `sortedChildren` is not being passed into the component.

var-kyle (changes requested):
Looks great, just one little change.

will0684 (approved):
Looks and works great on desktop and mobile great work @shawn320 and @var-kyle

var-kyle (approved):
Awesome, great job!

PR #306, opened 2021-07-29
UseEffect change

EphraimAssi:
# Description

[#380 Make sure page loads are being tracked after the first page load for analytics](https://trello.com/c/Z8tmWpF0/380-380-make-sure-page-loads-are-being-tracked-after-the-first-page-load-for-analytics)

The previous page load is now working but the useEffect I used for the page load seems to be running multiple times so it's messing up the results. I added a empty parameter to make it run only once [(https://css-tricks.com/run-useeffect-only-once/)](https://css-tricks.com/run-useeffect-only-once/)

## Test Instructions

1.Check to see if every page has an empty parameter in the useEffect for the page load.

## Checklist

## Product and Sprint Backlog


will0684 (approved):
Awesome work 👍

PR #321, opened 2021-08-03
Add robots.txt

var-kyle:
# Description

[Add robots.txt](https://trello.com/c/3pfFtErP)

Add a dynamic robots.txt file by creating an robots api route and adding a rewrite in the config. This allows us to check the environment and return the correct robots.txt.

In dev, we don't want web crawlers scraping the site, so we deny all
```
User-agent: *
Disallow: /
```

While in production, we will allow scraping except for the api routes
```
User-agent: *
Disallow: /api/*
```

There is a new env variable called `ENVIRONMENT` that will either be `development` or `production` to indicate which environment it is.

I'm not sure if we need to add a sitemap since the site isn't very large, but I can add one if anyone thinks it would be necessary.

## Acceptance Criteria

Navigating to `/robots.txt` returns plain text in either format above.

## Test Instructions

1. Set your ENVIRONMENT env to either `development` or `production`
2. Navigate to `/robots.txt` and see either of the above text snippets

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog


will0684 (approved):
Nice work thanks!

PR #322, opened 2021-08-03
update README

will0684:
# Description

[Update README with clear instructions for setting up development](https://trello.com/c/uzFRlhW3/450-update-readme-to-include-clear-instructions-for-setting-up-local-dev-environment)

Updated README to include a little more clarity on enabling the various features during development.

## Acceptance Criteria

Instructions should be clear and easy to follow.

## Help Requested

- Anything you think could be added
- Should we add the values of the template ID's to the README? I feel like that would be safe to do.

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ NOTIFY_REPORT_A_PROBLEM_TEMPLATE_ID=ask for this
 
 ## Enabling Unsubscribe
 
+Required environment variables:
+
 ```code
 MONGO_URL=your mongo URL
 MONGO_DB=your mongodb name
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ npm i --legacy-peer-deps
 npm run dev
 ```
 
-Create and run a production build locally:
+Create and run a production build locally***:
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,39 +1,102 @@
...
+npm i --legacy-peer-deps
+npm run dev
+```
+
+Create and run a production build locally:
+
+```bash
+npm run build
+npm start
 ```
```
var-kyle, line 26: Might be a good idea to mention that all the envs need to be set before creating the build, as Next.js is pretty finicky about that.

var-kyle (changes requested):
This looks much better but there is maybe 1 thing that I think would help new devs.

README.md
will0684: Do you mean as in the build fails if certain ones are not set? I am not sure what those are as I just tested it with a couple client-side and server-side ones and didn't have any failed build.

var-kyle: Right, you did specify that the CMS api url is required. I just thought a warning that the build will fail if you don't have the necessary envs would help. I'll approve since you do mention the required env though.

var-kyle (approved):
Nice work!

README.md
will0684: @var-kyle done!

var-kyle (approved):
Thank you, looks great and is clearer for sure!

PR #326, opened 2021-08-03
report a problem email fixes

will0684:
# Description

[#462 Report a problem should indicate the type of problem submitted](https://trello.com/c/yOdOOCCD/462-462-report-a-problem-should-indicate-the-type-of-problem-submitted)

[#461 Report a Problem feedback should indicate which page the user was on when they submitted the feedback](https://trello.com/c/M756GpH9/461-461-report-a-problem-feedback-should-indicate-which-page-the-user-was-on-when-they-submitted-the-feedback)

There were two issues with the format of the Report a Problem feedback emails:

1. Ashley Evans requested that the issue type be indicated in the subject line of the email. The app has been configured to pass the values that GCNotify expects to be able to render these conditionally.

2. The Report a Problem emails only indicated the root of the page from which the feedback is being sent when it should include the subpath. Our referrer-policy has been changed to be able to pass the URL with subpath into the report a problem API call.

## Acceptance Criteria

- Problem-type should be indicated in the subject line of the email.
- The specific page that the user sends the feedback from should be indicated in the email.

## Test Instructions

1. Use an email you have access to as the value for `NEXT_PUBLIC_NOTIFY_REPORT_A_PROBLEM_EMAIL` 
2. Try different selections in the Report a Problem component
3. See that specific page and problem type are included in the email

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add robots.txt
 
+## Fixes
+
+- Fixed bug with Report a Problem not sending along certain selection options
+
 ## [v1.0.3] - 2021-07-30
 
 - Modified Digital Center page content
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add robots.txt
 
-## Fixes
+## Fixed
 
 - Fixed bug with Report a Problem not sending along certain selection options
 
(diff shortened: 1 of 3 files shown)


var-kyle (commented):
The subject line in the email needs better formatting when multiple types are selected. I'm not sure Notify can handle a somewhat complex subject line, so it might be worth creating it in the code and passing it in to the Notify api.

will0684:
> The subject line in the email needs better formatting when multiple types are selected. I'm not sure Notify can handle a somewhat complex subject line, so it might be worth creating it in the code and passing it in to the Notify api.

Good call, an oversight on my part. Taking a look at things now.

will0684:
Ok @var-kyle changed the template and this is what it looks like with multiple boxes checked. Let me know what you think.

image

var-kyle (approved):
Awesome work! The email looks great!

PR #327, opened 2021-08-03
Fix build issue

var-kyle:
# Description

The pipeline is failing to build the application. I found that the package `next-page-tester` was at fault, so I've updated it to it's latest version (which supports Nextjs 11), but it was also in the production dependencies when it shouldn't need to be there, so I removed it from there as well.

## Product and Sprint Backlog


will0684 (approved):
Was trying some things out, but this is the obvious answer if it is compatible with Next 11. Sorry for the screw up 🤦

PR #330, opened 2021-08-04
Report a problem update

shawn320:
# Description
https://trello.com/c/W6wQcxMm/441-report-a-problem-list-items-would-not-be-announced-by-a-screen-reader-in-form-mode

- Move the bullet list and the "privacy statement" link to the top of the form so the screen reader will read the description first
- Add a title, and modify the contents in the bullet according to the new design
- The bullet list and the link should now live outside of the form

## Test Instructions

- Visit https://alphasite.dts-stn.com, scroll down to the bottom of the page, click on "Report a problem or a mistake on this page" button 
- Check if the change matches the design for both English and French

## Checklist

- [ ] Update CHANGELOG


components/organisms/ReportAProblem.js
```diff
@@ -268,48 +83,220 @@ export function ReportAProblem(props) {
...
-            type="submit"
-            dataCy="report-a-problem-submit"
-            dataTestId="report-a-problem-submit"
-            analyticsTracking
+          <div className="mb-6">
+            <a
+              className="underline text-xs sm:text-sm font-body hover:text-canada-footer-hover-font-blue text-canada-footer-font"
+              href={t("reportAProblemPrivacyStatementLink", {
+                lng: props.language,
+              })}
```
var-kyle, line 97: We have our own privacy statement page, should this link get changed to that instead on canada.ca's privacy statement?

var-kyle (changes requested):
This looks good, I think we should link to our own privacy statement though.

components/organisms/ReportAProblem.js
shawn320: > We have our own privacy statement page, should this link get changed to that instead on canada.ca's privacy statement?

That's a good question, let me confirm with the design team.

shawn320: The privacy link is now using the one on Alpha site

var-kyle (approved):
Nice work! Looks good and matches the design. 💯

PR #333, opened 2021-08-05
Analytics fix

EphraimAssi:
# Description

[#465 Fix small analytics issues](https://trello.com/c/xzGBglDR/465-465-fix-small-analytics-issues)

Fix incorrect page titles and adjust the timing of when the beacon is called so that the current page title is captured from the metadata and not the previous page. The change I made for the page load seems to have caused another issue so I'm reverting it to what it was.

## Test Instructions

- 1. Go on the splash page
- 2. Select a language to go on the home page
- 3. Write "adobeDataLayer" in the console, and the array should have only two objects.
- 4. The meta tag with the name "dcterms.title" should all be in this order (Name of the page - Service Canada Labs)

## Checklist

- [ ] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/pages/404.js
+++ b/pages/404.js
@@ -4,9 +4,17 @@ import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 import Link from "next/link";
 import { ReportAProblem } from "../components/organisms/ReportAProblem";
 import { ActionButton } from "../components/atoms/ActionButton";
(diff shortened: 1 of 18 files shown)


will0684 (approved):
Nice work, hopefully we're getting closer.

Issue #231, opened 2021-07-21
Missing headings in the header and footer document sections (WCAG 1.3.1 failure)


neilmispelaar:
### Description 

Each document section requires heading structure to indicate to the user programatically what the section is about or what content that they can expect to find in the section. 

At present the `header` and `footer` sections of the document are missing headings. 

It could be as simple as "Site header" and "Site footer", and it can be hidden, but there must be an `h2` heading in the document section. 

Furthermore, if the document section has subsections, it's important to add headings to those sections as well. 

For example, the `header` structure could be broken down to something like this: 

```
<header>
  <h2 class="sr-only">Service Canada Labs site header</h2>
  
  <section>
    <h3 class="sr-only">Test Site Notice</h3>
   ....
  </section>

  <section>
    <h3 class="sr-only">Language toggle</h3>
   ....
  </section>
  
  <nav>
    <h3 class="sr-only">Main site navigation</h3>
    <ul>....</ul>
  </nav>

</header>
```

### WCAG Resources 
https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H69

### Screenshots

<img width="1699" alt="Screen Shot 2021-07-21 at 1 56 19 PM" src="https://user-images.githubusercontent.com/9325038/126536933-c9b00b5c-e1e4-4aa4-9a2c-1158d311e81e.png">

<img width="1699" alt="Screen Shot 2021-07-21 at 1 58 02 PM" src="https://user-images.githubusercontent.com/9325038/126537273-7798ddbe-c947-4b23-9969-cb67e7589125.png">

PR #335, opened 2021-08-09
Add heading in contents for screen reader

shawn320:
# Description
This pr fixes the issue #231

Added headings for screen reader. Added ```<h2>``` heading in Header and Footer and ```<h3>``` headings in the navigations. The heading will not display in the browser but will be announced by the screen reader.

Heading titles and French translation:
Screen Shot 2021-08-16 at 2 53 17 PM


components/organisms/Layout.js
```diff
@@ -43,18 +43,21 @@ export const Layout = ({
         </a>
       </nav>
       <header>
+        <h2 className="sr-only">Main site header</h2>
```
will0684, line 46: Nice work, I think these strings should be bilingual

will0684 (changes requested):

components/organisms/Layout.js
will0684: Message Isabelle Waltzing with what you would like translated, she is probably best to provide them.

shawn320: Thank you, I'll message her for the translate.

will0684 (changes requested):
Waiting on review from Isabelle

will0684 (approved):
Good stuff thanks Shawn 👍

PR #337, opened 2021-08-10
Anonymous feedback

EphraimAssi:
# Description

[#472 Create anonymous feedback component](https://trello.com/c/X9ywObnp/472-472-create-anonymous-feedback-component)

I did the feedback component and linked it to the phase banner. It doesn't submit anything yet.
image
image
image
image

## Acceptance Criteria

It should like the image that is in this task on trello (https://trello.com/c/G3pLR1BW/29-29-qualitative-product-owner-feedback)

## Test Instructions

1. There isn't any active experiments yet, so try it in storybook or in the layout component, go to the Feedback and change the active prop for true.
2. Check how it looks like on big and small screens compared to the picture here (https://trello.com/c/G3pLR1BW/29-29-qualitative-product-owner-feedback)

## Help Requested

- Things that you (the submitter)...
- want reviewers to pay very close...
- attention to when they review this PR.

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog


components/atoms/PhaseBanner.js
```diff
@@ -1,22 +1,86 @@
...
+    feedbackDropdown.classList.toggle("active");
+    feedbackButton.getAttribute("aria-expanded") === "true"
+      ? feedbackButton.setAttribute("aria-expanded", false)
+      : feedbackButton.setAttribute("aria-expanded", true);
+
+    feedbackButton.getAttribute("aria-expanded") === "true"
+      ? (caret.innerHTML = "&#9660;")
+      : (caret.innerHTML = "&#9656;");
+  }
+
```
var-kyle, line 29: I would like to see this handled with state, since this is a React app.

components/organisms/Feedback.js
```diff
@@ -0,0 +1,188 @@
...
+
+    feedbackDropdown.classList.toggle("active");
+    feedbackButton.getAttribute("aria-expanded") === "true"
+      ? feedbackButton.setAttribute("aria-expanded", false)
+      : feedbackButton.setAttribute("aria-expanded", true);
+
+    feedbackButton.getAttribute("aria-expanded") === "true"
+      ? (caret.innerHTML = "&#9660;")
+      : (caret.innerHTML = "&#9656;");
+  }
```
var-kyle, line 52: Similar to the submit feedback function, I think this should be handled using React state.

components/organisms/Layout.js
```diff
@@ -25,6 +25,7 @@ export const Layout = ({
   locale,
   langUrl,
   breadcrumbItems,
+  active,
```
var-kyle, line 28: This should be more descriptive, no one would know what "active" means just by looking at it.

var-kyle (changes requested):
The component looks great on desktop and mobile, great work with the styling!

There are a few things that need a little work though. The form should use the JOI package for validation rather than just the default browser implementation - see the signup page to see how it's been done. Some of the state changes should use react state.

components/atoms/PhaseBanner.js
```diff
@@ -1,22 +1,86 @@
...
+                aria-expanded="false"
+                aria-controls="feedbackDropdown"
+                data-testid="feedbackButton"
+              >
+                <span id="caret" className="text-p leading-5">
+                  &#9656;
+                </span>
+                <span className="underline mt-2 lg:mt-0 ml-2 font-bold">
+                  {t("giveFeedback")}
+                </span>
```
pcraig3, line 74: We've gotten back a11y audits from the accessibility office a few times saying we should use `<strong>` instead of the "font-bold" class to more clearly convey emphasis.

components/organisms/Feedback.js
```diff
@@ -0,0 +1,188 @@
...
+                &times;
+              </span>
+              <span className="text-xs leading-4 lg:text-sm underline ml-2 lg:leading-10">
+                {t("close")}
+              </span>
+            </button>
+          </div>
+          <ul className="list-outside list-disc px-6 py-2">
+            <li className="text-xs lg:text-sm font-body mb-4">
+              <b>{t("reportAProblemNoReply", { lng: props.language })}</b>{" "}
```
pcraig3, line 120: We should usually pick `<strong>` instead of `<b>`, although the difference is subtle.

- If the bolded text is worth emphasizing (eg, "you really have to pay attention this"), then we should use `<strong>`
- If the bolded text is just for looking bold on the page to fit in with the design, then `<b>` is good.

In this case, I think we do want to emphasize that people won't receive a reply.

👍

components/organisms/Feedback.js
```diff
@@ -0,0 +1,188 @@
...
+            <p className="text-xs lg:text-sm my-2">{t("maximum2000")}</p>
+            <textarea
+              id="feedbackTextArea"
+              name="feedbackTextArea"
+              maxLength="2000"
+              rows="5"
+              className={
+                "text-input font-body w-full min-h-40px shadow-sm text-form-input-gray border-2 py-6px px-12px rounded"
+              }
+              required
```
pcraig3, line 165: "required" means that the browser validation takes over, which we tend to get rid of and handle using our error messages. We should remove this and use the error message / validation message pattern we already have for the signup page.

pcraig3 (changes requested):
I did a bit of a visual review here @EphraimAssi, and the inside of the component is good, but the "open/close" portion of it is a bit wonky.

### Fixing the "open/close triangle" 

The caret is the most noticeable part here. When you expand the feedback widget, it gets way bigger -- I think because it's not a set width but just expands to be as big as the container. Here's a gif where you can see it changing size.

Screen Recording 2021-08-11 at 12 34 13

Also, on mobile, you can see that the caret doesn't line up with the text anymore.

| before | after |
|--------|-------|
|  when collapsed, the little arrow should point at the center of "Give"       |  When opened and focused, you can see that it overlaps the text above     |
|  Screen Shot 2021-08-11 at 12 35 03  |   Screen Shot 2021-08-11 at 11 07 22  |

Pretty much, what we're looking for is something more like the little triangle on the "report a problem" widget. I would suggest using the flexbox "baseline" horizontal alignment for aligning the triangle with the words.

### The feedback widget itself

The title and particularly the "close " button are too close to the area above them on tablet sizes. I would put more top margin to give a bit more space between the feedback widget itself and the beta banner.

Screen Shot 2021-08-11 at 12 35 21

components/organisms/PhaseBanner.js
```diff
@@ -0,0 +1,297 @@
...
+
+  //Function for changing feedback button state
+  function onfeedbackClick() {
+    const feedbackDropdown = document.getElementById("feedbackDropdown");
+
+    feedbackDropdown.classList.toggle("active");
+    ariaExpanded === true ? setAriaExpanded(false) : setAriaExpanded(true);
+
+    ariaExpanded === false ? setCaretHtml("&#9660;") : setCaretHtml("&#9656;");
+  }
```
var-kyle, line 89: I'm still not sure I like how this is being done, but we can discuss it. You're using aria attributes and I'm not super familiar with them. I would have done this similar to how the ReportAProblem component is done though, where it's either on the page, or not there are all, instead of toggling visibility. Of course, if there are a11y issues with the way the ReportAProblem component is doing it, then it makes sense not to do it that way.

I would remove this function entirely, as well as removing `ariaExpanded` and `caretHtml`, and have `showFeedback` instead.

components/organisms/PhaseBanner.js
```diff
@@ -0,0 +1,297 @@
...
+                aria-controls="feedbackDropdown"
+                data-testid="feedbackButton"
+              >
+                <span
+                  id="caret"
+                  className={`${
+                    caretHtml === "&#9660;" ? "text-sm" : "text-p"
+                  } leading-7 lg:leading-5`}
+                  dangerouslySetInnerHTML={{ __html: caretHtml }}
+                />
```
var-kyle, line 136: If the `onFeedbackClick()` function is removed and the `showFeedback` state is added, this could then look like:
```
<span 
    id="caret" 
    className={`${showFeedback ? "text-sm" : "text-p"} leading-7 lg:leading-5`}
    dangerouslySetInnerHTML={{ __html: showFeedback ? "&#9660;" : "&#9656;" }}
/>
```

components/organisms/PhaseBanner.js
```diff
@@ -0,0 +1,297 @@
...
+              ""
+            )}
+          </div>
+        </div>
+      </div>
+      <div
+        id="feedbackDropdown"
+        className="feedbackDropdown bg-custom-blue-blue"
+        data-testid="feedbackDropdown"
+      >
```
var-kyle, line 151: This `div` would be wrapped by `showFeedback`,
```
{showFeedback ? (
        <div
          id="feedbackDropdown"
          className="bg-custom-blue-blue"
          data-testid="feedbackDropdown"
        >
...
</div>
: ""}
```
I'm not sure if this div should always be rendered or not since you're using `aria-control` to specify that the button will open or close this div.

components/organisms/PhaseBanner.js
```diff
@@ -0,0 +1,297 @@
...
+              </h2>
+              <button
+                id="feedbackClose"
+                onClick={onfeedbackClick}
+                className="font-body text-white flex mt-2.5 lg:mt-0"
+                aria-haspopup="true"
+                aria-expanded={ariaExpanded}
+                aria-controls="feedbackDropdown"
+                data-testid="closeButton"
+              >
```
var-kyle, line 166: This can look like the open button.

components/organisms/PhaseBanner.js
```diff
@@ -0,0 +1,297 @@
...
```
var-kyle, line 198: This close button would look the same as the open button.

components/organisms/PhaseBanner.js
```diff
@@ -0,0 +1,297 @@
...
+            {feedbackActive ? (
+              <button
+                id="feedbackButton"
+                onClick={onfeedbackClick}
+                className="bg-circle-color font-body text-xs lg:text-sm text-white flex text-left sm:ml-4 my-2 sm:my-0"
+                aria-haspopup="true"
+                aria-expanded={ariaExpanded}
+                aria-controls="feedbackDropdown"
+                data-testid="feedbackButton"
+              >
```
var-kyle, line 129: The `onClick` function in the button can look like this:
```
onClick={() => {setShowFeedback(!showFeedback);}}
```
The `aria-expanded` attribute can use the `showFeedback` variable too, since if `showFeedback` is true, it's expanded and if it's false, it isn't expanded.
```
aria-expanded={showFeedback}
```

styles/globals.css
```diff
@@ -167,4 +167,12 @@ html {
...
 }
 
+.active {
+  display: block;
+}
+
+.feedbackDropdown {
+  display: none;
+}
+
```
var-kyle, line 177: These wouldn't be necessary if you can change the way the feedback works.

var-kyle (changes requested):
A lot of good work here, but I would prefer to see it behave more like the ReportAProblem component with showing it's contents.

EphraimAssi:
I removed all the aria attributes since apparently it's more for suitable for a menu.

CHANGELOG.md
```diff
@@ -7,12 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+<<<<<<< HEAD
+- Updated PhaseBanner to include Feedback component
+=======
 ## [v1.0.6] - 2021-08-13
 
 ## Changed
 
 - Updated `/signup`, `/thankyou`, `/confirmation`, and `/unsubscribe` pages with the proper content and translations
 - Added button on Virtual Assistant page linking to Virtual Assistant prototype
+>>>>>>> main
```
var-kyle, line 19: Looks like this didn't get merged properly.

components/organisms/PhaseBanner.js
```diff
@@ -0,0 +1,283 @@
...
+            ) : (
+              ""
+            )}
+          </div>
+        </div>
+      </div>
+      {showFeedback ? (
+        <div
+          id="feedbackDropdown"
+          className="feedbackDropdown bg-custom-blue-blue"
```
var-kyle, line 138: The `feedbackDropdown` class doesn't do anything, it should be ok to remove it.

var-kyle (changes requested):
Looks really good, just a couple of little things before I think it's good to go!

EphraimAssi:
Added error message and design retouches

components/organisms/Layout.js
```diff
@@ -45,7 +46,7 @@ export const Layout = ({
       <header>
         <h2 className="sr-only">{t("globalHeader")}</h2>
         <h3 className="sr-only">{t("testSiteNotice")}</h3>
-        <PhaseBanner phase={t("phaseBannerTag")}>
+        <PhaseBanner phase={t("phaseBannerTag")} feedbackActive={true}>
```
var-kyle, line 49: I don't think we want the feedback active on our site.

var-kyle (changes requested):
I think this is looking really good.

Apart from having the feedback active on our site, there is an issue with the border when tabbing on to the expand button. The border should probably not be black as it's hard to see that focus is on it.
image

var-kyle (changes requested):
The expand button looks much better now, but I didn't check the other links/buttons, so I quickly did that and the circled things in the image below could also use a focus border update.

image

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Updated PhaseBanner to include Feedback component
```
var-kyle, line 10: Please move this under an appropriate header. Either `## Added` or `## Changed` should be fine

var-kyle (approved):
Awesome work, just one last little change and it should be ready to go!

var-kyle (approved):
Awesome work @EphraimAssi , thank you!

PR #341, opened 2021-08-11
External link icon

shawn320:
# Description
https://trello.com/c/lJuvAKUF/478-478-add-external-link-icon-to-project-cards-titles-that-send-ppl-directly-to-the-prototype

Add an external link icon to the experiment card. Experiment card that has a full url will show the icon beside the title. Experiment card that has a relative url will not have the icon.  

Screen Shot 2021-08-11 at 12 16 55 PM


var-kyle (changes requested):
Mostly looks good, I would just don't like that there is another property being added for essentially no reason.

components/molecules/Experiment.js
```diff
@@ -22,10 +22,13 @@ export const Experiment = (props) => {
     >
       <Link href={props.href}>
         <a
-          className="block text-p text-custom-blue-projects-link underline hover:opacity-70"
+          className="flex block text-p text-custom-blue-projects-link underline hover:opacity-70"
           tabIndex="0"
         >
           {props.title}
+          {props.url && props.url.substring(0, 8) === "https://" ? (
```
will0684, line 29: Instead of adding another property, would it not be possible to just use the href prop that is already being passed in? Regardless of language the protocol should be the same.

will0684 (changes requested):

components/molecules/Experiment.js
shawn320: Yes, you are right. Thank you for point it out!

will0684 (approved):
LGTM on desktop and mobile, nice work 👍

Issue #228, opened 2021-07-21
A11y issues identified by Sidra


pcraig3:
Hey everyone, @sidramatik did a lil WCAG lookover, and here's what she found:

- [x] I think the menu links need to be underlined because they look a lot like normal text	
- [x] Focus state on breadcrumb is showing inconsistent borders
- [x] Menu needs a close button (even sr-only is okay) and a state change to indicate menu closing
-  [x] That alert on the very top that lets you know it’s a test site should maybe be enclosed as an `<alert or with role=alert>`
- [x] There are 4 navigation levels (primary, breadcrumb, and two standard footer navs) and they each should be labeled as such: https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/landmarks/navigation.html
-  [x] Footer passed contrast checks but give a dark bg color to .footerBackground so it doesn’t get flagged by automatic contrast checkers (the image there passes but it’ll get flagged)
- [x] “Sign up to get invited to research…” pre-footer needs a role to indicate that it isn’t primary content i.e. directly related to SC labs.
- [x] The "Sign up to get invited to research sessions" button doesn’t pass contrast check - can we switch the active state to a white outline (2px thick)?

I've made them all into checkboxes so that we can check them off as we go. Some of these issues are design-related (eg, the Figma designs don't have underlines under the menu links), but not all of them. We should respond to them one by one and close them off. (We can also break individual bullets into their own issue if they are big and thorny.)

sidramatik:
*Correction, there are 4 levels of navigation including the breadcrumb (primary, breadcrumb, and the two standard footer navs)

pcraig3:
Added an `<aside>` to the signup section at the bottom so that it's indicated as not primary content.

PR #352, opened 2021-08-17
Navigation Landmark

shawn320:
# Description

#228

There are 4 navigation levels (primary, breadcrumb, and two standard footer navs) and they each should be labeled as such: https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/landmarks/navigation.html

## Acceptance Criteria

Add ARIA landmark to each navigation. The screen reader will anounce the header, and let the user know if they are at navigation links


will0684 (approved):
Thanks Shawn!

PR #353, opened 2021-08-19
Sign up button contrast

shawn320:
# Description

The constrast ratio between the dark blue background and the blue sign-up button near the bottom of the page is 1.58:1 (needs to be 3:1 or higher).

https://trello.com/c/etJ3bws8/514-button-contrast-fix

Adjust the button's background color to have at least 1:3 ratio with the background. 

Screen Shot 2021-08-24 at 12 25 11 PM
Screen Shot 2021-08-24 at 12 26 07 PM


tailwind.config.js
```diff
@@ -86,7 +87,7 @@ module.exports = {
         "text-gray": "#5c5c5c",
         "hr-red-bar": "#AE3B43",
         "canada-ca-link-colour": "#7834BC",
-        "circle-color": "#083C6C",
+        "circle-color": "#062F56",
```
var-kyle, line 90: Was there a reason to change this one too? The Figma designs still show `#083C6C` for the background colour for the phase banner and the CallToAction component.

shawn320: I made it slightly darker because the original one will require a lighter button color to meet the required contrast ratio, but will not pass the contrast check again because not enough contrast with the white button text.

components/molecules/CallToAction.js
```diff
@@ -27,7 +27,7 @@ export function CallToAction(props) {
                 <ActionButton
                   id="become-a-participant-btn"
                   custom={
-                    "py-2 px-4 bg-custom-blue-blue text-white border border-custom-blue-blue active:bg-custom-blue-dark active:border-2 active:border-white hover:border-2 hover:border-white hover:bg-custom-blue-light"
+                    "py-2 px-4 text-custom-blue-projects-link bg-details-button-gray hover:bg-gray-300 border border-custom-blue-blue active:bg-custom-blue-dark active:border-2 active:border-white hover:border-2 hover:border-white"
```
var-kyle, line 30: Can you add `rounded` in here? I think most other buttons have rounded corners and this one looks a little off without it.

var-kyle (changes requested):
The button looks better, but it's just a little off with the corners.

var-kyle (approved):
Awesome, looks great!

PR #355, opened 2021-08-20
Language detection and redirects

var-kyle:
# Description

[# 486 You always land on the splash page – in the YDG you only land there if the system cant figure out your preferred language.](https://trello.com/c/UGWaEKfH)

Updated: fixed image not loading issue.

This can be up for discussion, but this seems like it works well.

### How this works

When a new user visits the site (the index page, not the other pages) they are given the choice to see the site in English or French. When they make their choice we set a client-side cookie (using the [cookie-cutter](https://www.npmjs.com/package/cookie-cutter) npm package) called `lang`:
```
const setLanguage = (event) => {
  event.target.id === "french-button"
    ? cookieCutter.set("lang", "fr")
    : cookieCutter.set("lang", "en");
};
```

With this cookie set, we can now perform redirects to the correct language on any page. For example, you've chosen to see the site in French when you first visited and that language selection is now stored in the `lang` cookie. Then you leave the site. Now if you get an email with a link to some page on the site (lets say the link was `/projects`, not `/fr/projects`, when you clicked the link we would see that the cookie says you wanted French so we perform a redirect to `/fr/projects`.

Similarly, the language toggle link in the Layout component will update the cookie so that it will always remember what language you wanted to see the site in.

### Possible issue?

Last time we had this done, language selection was stored in local storage. I've chosen to use a cookie here because I can take advantage of [Next.js redirects](https://nextjs.org/docs/api-reference/next.config.js/redirects) in nextjs config. If there was a choice to not use cookies made at some point, we can change this to use local storage again, but we won't be able to use the redirects built into Next.js.

## Acceptance Criteria

After setting your language of choice, pages load in your specified language, regardless of URL entered.

## Test Instructions

1. Navigate to `/`
2. Pick your language and notice a cookie called `lang` is created (by looking in your browsers dev tools)
3. Try navigating to a page in the other language and notice a redirect to the page in the language you've selected

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog


will0684 (dismissed):
Tested and looks good! Great work 👍

will0684 (approved):

var-kyle (commented):
There is still something funny going on here. The splash page loads fine on first load, but once the cookie is set, images stop loading for me:
image

shawn320:
> There is still something funny going on here. The splash page loads fine on first load, but once the cookie is set, images stop loading for me: image
This is weird, it works fine on my end. Did you get any errors in the console? What browser did you use?

shawn320 (approved):

PR #356, opened 2021-08-23
Home banner contrast

shawn320:
# Description

-1.4.3 Contrast (Minimum): Setting the browser viewport width to anywhere between 450-766 pixels causes the banner’s white text to appear over the yellow lightbulb... with a contrast ratio of 2.94:1 (needs to be 4.5:1 or higher). https://trello.com/c/ppU6Uh7d/516-homepage-banner-contrast

**How to fix:** Adjust the background opacity to 30% for smaller screen.

**Before:**
Screen Shot 2021-08-23 at 11 03 08 AM

**After:**
Screen Shot 2021-08-23 at 11 03 52 AM


var-kyle (changes requested):
The contrast looks good to me. Just don't forget to update the changelog with the changes you've done.

var-kyle (approved):
Awesome, thank you!

PR #368, opened 2021-08-26
Sr only close button for menu

EphraimAssi:
# Description

Updated menu to have a close sr-only close button under 768px and added a state change to indicate menu closing. Also changed the onmenuClick function to work with a state now.
#228 

## Acceptance Criteria

Menu should have a close button when screen is under 768px width. The "menuDropwdown" class should have an aria-expanded indicating that it is closed or open.

## Test Instructions

1. Run the code locally and inspect put the screen width under 768px
2. Open the menu and tab to the close button
3. Check to see if the close button is invisible when not tabbed, and if the aria-expanded state changes in the console onClick.
4. Put the screen over 768px and try tabbing to the close button (It shouldn't work, if it does, it's now well implemented)

## Checklist

- [ ] Update CHANGELOG

## Product and Sprint Backlog


components/molecules/Menu.js
```diff
@@ -31,22 +22,45 @@ export function Menu(props) {
...
+          className={`${
+            showMenu ? "" : "hidden"
+          } sr-only font-body flex text-canada-footer-font outline-none focus:not-sr-only focus:outline-black-solid lg:invisible`}
+          aria-expanded={showMenu}
+          aria-controls="menuDropdown"
+          aria-label="Close"
+          data-testid="menuCloseButton"
+        >
+          <span id="close" className="text-h2">
+            &times;
```
will0684, line 53: I think we got some feedback in the feedback component a11y audit that says we should avoid using this format for the close button span value. I think it's because the screen reader will actually read out "times close button" when it is looking at the close menu button. More detail on their feedback is in #6 here: https://docs.google.com/spreadsheets/d/1HY1tHL6YTsapS0jCcD5w6hm9uIfFCHIk8AHC4ITY1wg/edit#gid=0

EphraimAssi:
Updated this pr to include update for "Providing descriptive labels" for the feedback component since both issues are similar

EphraimAssi:
I'm not sure why the testing is failing.

PR #377, opened 2021-09-13
Translate pathname for French route

shawn320:
# Description

[French subpath shoud be in French](https://trello.com/c/j3vICOBi/515-french-subpaths-should-be-in-french-language)

Pathname in url should be translate to French when switch to French route. For example, the French route should be `/fr/merci` instead of `/fr/thankyou`. 

Next js has a method named [rewrite](https://nextjs.org/docs/api-reference/next.config.js/rewrites), which allow you to rewrite path to a specific page. This method is used in this pr to handle the translated route. 

## Acceptance Criteria

When you switch the page contents to French, the pathname in url should be translated to French as well.

## Test Instructions

Swith language to French on any page.


will0684 (approved):
Really great work here Shawn. Tested and working well 🚀

var-kyle (commented):
I still find it a little buggy, for instance, when you go to `/accueil` you get redirected to `/home` and you need to put the `/fr/` first to actually get to the french page. But that makes things kind of confusing for users, since I can't imagine they'll know to add that `/fr/` first.

There are also a few links that still have the english urls (eg. on the splash page, the french link is `/fr/home` rather than `/accueil`).

will0684:
Ugh thanks @var-kyle I really was deficient in my testing of this obviously...sorry about that. Regardless good work so far @shawn320

pages/about.js
```diff
@@ -2,16 +2,19 @@ import { Layout } from "../components/organisms/Layout";
...
 
 export default function About(props) {
   const { t } = useTranslation("common");
-  const { asPath } = useRouter();
+  const router = useRouter();
 
   useEffect(() => {
+    if (window.location.pathname === "/a-propos") {
+      router.push("/fr/a-propos");
+    }
```
var-kyle, line 17: I really don't think this is a good idea, it would result in some people seeing the english page load, then disappear and  then the french page would show up - not a great user experience.

To fix this, in `ActionButton`, you can simply add `locale={props.lang}` in the `<Link>` and it will automatically do the language changing for you.

var-kyle (dismissed):
As much as I would like the `/fr/` to be removed from the url, I don't see a way to get the i18n package to pick up what language we have set at the moment, so it's fine to leave it, but there are a couple of things that I would like to see changed (mostly just removing the `router.push()`).

components/organisms/Layout.js
```diff
@@ -110,15 +110,15 @@ export const Layout = ({
             signUpText={t("signupBtn")}
             items={[
               {
-                link: "/projects",
+                link: locale === "fr" ? "/projets" : "/projects",
```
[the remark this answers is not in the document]
shawn320: Good idea! It's much cleaner!

will0684 (dismissed):
Great work here @shawn320 this will make a lot of people happy :)

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Changed Digital Center URL from 'projects/digital-center' to 'projects/digital-centre'
 - Updated breadcrumbs on the privacy page
 - Updated project status label colours
+- Updated French path in French language
```
will0684, line 39: This should be under a ##Changed header under the ##Unreleased header

components/atoms/ActionButton.js
```diff
@@ -56,6 +56,7 @@ export function ActionButton(props) {
         role="button"
         draggable="false"
         lang={props.lang}
+        locale={props.langUrl}
```
var-kyle, line 59: What is the purpose of `props.langUrl`?

var-kyle (commented):
Definitely looks a lot nicer and behaves well, I'm just a little confused about a property in the ActionButton.

components/atoms/ActionButton.js
shawn320: It definitely doesn't do anything, It's not used in any of the pages, I'm removing it.

will0684 (dismissed):
It's looking good to me 👍

will0684 (approved):

PR #386, opened 2021-09-22
Update purge safelist in tailwind config

var-kyle:
# Description

Since the project status label colours are being added dynamically, the purge function that tailwind uses didn't capture those classes and removed them, resulting in colourless status labels:

image

This PR adds those css classes to the safelist so that they aren't removed when building the application, ensuring that the colours remain:

image

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog


will0684 (approved):
Nice work!

PR #397, opened 2021-10-04
Feedback a11y fixes

krischarbonneau:
# Description

[#569 Address PhaseBanner/Feedback component issues raised in the a11y audit](https://trello.com/c/qHCQE31w)

Took care of several of the a11y issues outlined in the task linked above. These include: 

- I've made it more clear that the text area is a required field
- I've added a function that makes the screen reader announce whatever phrase is passed to it. Using this, the screen reader now announces the state of the feedback form (whether it is expanded or collapsed) when clicked. The state is also announced when a keyboard user tabs to the button element
- Using the same function as in the above task, the error message that appears whenever a user submits an empty feedback form is now announced by the screen reader
- The two `<p>` tags above the feedback form textarea are now announced to screen reader users by use of `aria-describedby`
- The feedback form close button is now before the `<h2>` in page order

## Test Instructions

1. Pull in branch
2. Change `feedbackActive={feedbackActive}` to `feedbackActive={true}` in `Layout.js`
3. Type `npm run dev`
4. Use your favorite screen reader to confirm the above changes/fixes work as expected

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Updated content on the experiments, about and sign up pages
 - Updated French path in French language
+- Added text to the feedback form to clearly indicate that the text area is required
+
+## Fixed
+
+- Screen reader users are now notified of the current state of the feedback form (whether it's expanded or collapsed)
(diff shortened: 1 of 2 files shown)


shawn320 (approved):
Tested with screen reader, everything works well! Nice work!

Issue #142, opened 2021-06-27
Page title on the "experiments" page is cut off on mobile phones


pcraig3:
Pretty small issue here: the H1 is cut off on the "experiments" page when I check on my phone.

<img width="500" src="https://user-images.githubusercontent.com/2454380/123556904-cd391c80-d75b-11eb-91e6-dd4610df0d57.png" />

PR #404, opened 2021-10-14
E2E test signup and unsubscribe workflow

shawn320:
# Description

[#142 Write e2e test for signup and unsubscribe workflow](https://jira-dev.bdm-dev.dts-stn.com/secure/RapidBoard.jspa?rapidView=96&projectKey=SCL&view=detail&selectedIssue=SCL-142)

Test the signup and unsubscribe workflow. Use the `intercept` method to stub and mock a request's response. 

## Test Instructions

1. Start development server: `npm run dev`
2. Run the cypress test runnner in a new terminal: `npx cypress open`
3. Click on the `signup.spec.js` and `unsubscribe.spec.js` to run the tests

## Checklist

- [x] Update CHANGELOG

### DIFF
--- a/cypress/integration/signup.spec.js
+++ b/cypress/integration/signup.spec.js
@@ -130,6 +130,11 @@ describe("signup page", () => {
       cy.log(xhr);
       expect(xhr.request.method).to.equal("POST");
       expect(xhr.response.statusCode).to.equal(400);
+      cy.get('[data-cy="error-box"').should("exist");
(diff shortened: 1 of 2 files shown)


shawn320:
> These test are looking a lot better, but would you be able to add checks for what the page looks like to the new tests? Checking that the error message appears and has the message is what we expect would be ideal, instead of just checking the api response.

I see what you mean, thank you, I will add the tests that will redirect to the error page.

PR #406, opened 2021-10-18
Add metadata support

krischarbonneau:
# Description

[#608 All pages should support metadata for Twitter, OG, Dublin Core and HTML (Original Trello task)](https://trello.com/c/ngV9gSLb)
[#463 Add meta description to web pages](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-493)

This PR adds metadata support as per the original Trello task. I've added the meta descriptions based on the current translations from the Jira task.

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3. Go to any page and open the dev console
4. Expand `<head>` to see the newly added metadata tags

## Things of note

A few pages weren't included in the official translation doc, including:

- Error page
- 500 error page (404 was given a description but not 500)
- All of the project pages (DC, LJ, etc.)
- Splash page

I've added placeholders here for the time being but we should probably get design to write up some meta descriptions for these pages as well to be consistent.

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

[Jira](https://jira-dev.bdm-dev.dts-stn.com/secure/RapidBoard.jspa?rapidView=96)

### DIFF
--- a/pages/404.js
+++ b/pages/404.js
@@ -30,11 +30,11 @@ export default function error404(props) {
           <title data-gc-analytics-error="404">
             {`404 — ${t("siteTitle")}`}
           </title>
-          <meta name="title" content={`404 — ${t("siteTitle")}`} />
           <meta
             name="description"
             content={`${t("404errorMetaDescription")}`}
           />
+          <meta name="author" content="Service Canada" />
           <link rel="icon" href="/favicon.ico" />
 
           {/* DCMI Meta Tags */}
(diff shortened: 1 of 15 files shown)


krischarbonneau:
> Really great work here, should improve SEO greatly.
> 
> The only thing I see is that you've added `<meta name="title">...</meta>` and I don't think that's real metadata. Web crawlers will look at `<title>`, except for certain ones (like Facebook, Twitter, etc), but you've added those:
> 
> * dcterms.title
> * og:title
> * twitter:title

Yeah you're right. There's some stuff on StackOverflow about this being a meta tag but the consensus is that it's mostly ignored by most crawlers, Google included. I removed the title meta tag and added `<meta name="author" content="Service Canada" />`  in it's place as per the document provided by design. I didn't add `<meta name="keywords" content="[Keywords]" />` since they aren't currently included in the official translation doc.

PR #417, opened 2021-11-08
Ie not supported

krischarbonneau:
# Description

[DEV Handling IE not being supported](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-67)

This PR adds the `/notsupported` page which a user will be redirected to ONLY when they are using Internet Explorer. This is achieved by using the Redirects function in our next.config. Since most of our styling isn't compatible at all with IE (thanks Tailwind), I've also added a bunch of IE-only CSS to make it look the same. This comes with its quirks (see Help Requested section) but overall it works fine.

## Acceptance Criteria

When a user attempts to navigate to our site using Internet Explorer, they are redirected to a page telling them their browser isn't supported with information on how to download a supported browser.

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3. Launch Internet Explorer and navigate to `http://localhost:3000/`
4. You should be redirected to `/notsupported`
5. Click on the button that says `Copy link` (give it a second, IE seems to struggle but it does eventually copy the link to your clipboard) and paste it in your address bar to confirm the link was properly copied.
6. Click the links to make sure you're properly redirected to each browsers' download site

## Help Requested

Not that it's a huge issue since this page will only be presented to users on older browsers, but for whatever reason when the window is a certain size, there will be a very small amount of horizontal scroll. I assume there's some padding or margins doing some funky stuff somewhere but I haven't been able to spot where.

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

[Jira](https://jira-dev.bdm-dev.dts-stn.com/secure/RapidBoard.jspa?rapidView=96)


next.config.js
```diff
@@ -87,4 +88,20 @@ module.exports = {
...
+            type: "header",
+            key: "User-Agent",
+            value: "(.*Trident.*)",
+          },
+        ],
+        destination: "/notsupported",
+        permanent: false,
+      },
+    ];
+  },
```
var-kyle, line 106: This will only redirect you if you navigate to the splash page, it would be nice if the redirect happened from any page.

var-kyle (changes requested):
Nice work! The page looks good and the redirect is definitely the right way to handle this, but it would be nice if any page in the application can perform that redirect.

next.config.js
krischarbonneau: Given the fact that a catch-all regex pattern seems to prevent pages from rendering on IE (though it does redirect properly), I've added a rule for every page that checks if IE is being used. This is obviously not the most ideal since we'd need to add a new rule every time we add a new page, but it is the best we can do for the time being.

I also added IE detection to the 404 and 500 pages since they were not auto redirecting using the redirects function

krischarbonneau:
> Not sure if it's just me. The splash page works fine, but when I go to /home, or any other pages, the page won't load, and I'm getting this message in the console Capture .

I haven't been able to replicate this warning on my end, and the pages seem to be working fine, but worth looking into more...

var-kyle (approved):
Awesome work! Behaves as expected and looks good.

PR #429, opened 2021-11-16
Fix infinite redirect issue

shawn320:
# Description

[Infinite redirect bug when refresh page after switch language](https://jira-dev.bdm-dev.dts-stn.com/secure/RapidBoard.jspa?rapidView=96&projectKey=SCL&view=detail&selectedIssue=SCL-549)

**Context:**
After user selected prefered language from the splash page, if the language is switched again using the language button on top of the page, then refresh the page, the page will keep redirect back and force between English and French, causes an infinite redirect loop.

**What caused this:**
The [cookie-cutter](https://www.npmjs.com/package/cookie-cutter) package that we are using to set the cookie caused this issue. The `set` method will create a new cookie without overriding the previous one. For example, If user switch language again using the switch button on top of the page when the default language was already selected on the splash page, both `fr` and `en` will exist in the cookie with the same key, which confuses nextJS redirect. This package doesn't have the option to remove cookies that you previously set. 

**Solution:**
Found a package named [js-cookie](https://www.npmjs.com/package/js-cookie), this package is similar to `cookie-cutter`, but has the `remove` method that allow you to delete cookies. I ended up not using the `remove` method because the `set` method will override the previous cookie. 

## Test Instructions

1. pick a language on splash page
2. go to any page, click on the language button on top of the page
3. refresh the page to see if there are any issues
4. switch language and refresh again

## Checklist

- [x] Update CHANGELOG


components/organisms/Layout.js
```diff
@@ -31,6 +30,10 @@ export const Layout = ({
   const { t } = useTranslation("common");
   const language = locale === "en" ? "fr" : "en";
 
+  useEffect(() => {
+    console.log("lang: ", Cookies.get("lang"));
+  }, []);
+
```
var-kyle, line 36: Would you be able to remove this? We don't really want random console logs around the site.

pages/index.js
```diff
@@ -4,15 +4,15 @@ import { useTranslation } from "next-i18next";
...
 
 export default function Index(props) {
   const { t } = useTranslation("common");
 
   const setLanguage = (event) => {
     event.target.id === "french-button"
-      ? cookieCutter.set("lang", "fr")
-      : cookieCutter.set("lang", "en");
+      ? Cookies.set("lang", "fr")
+      : Cookies.set("lang", "en");
```
var-kyle, line 15: Can you add either the `secure` or `sameSite` option to the cookie?

components/organisms/Layout.js
```diff
@@ -8,12 +9,10 @@ import Link from "next/link";
...
 import { DateModified } from "../atoms/DateModified";
 import { Breadcrumb } from "../atoms/Breadcrumb";
-import cookieCutter from "cookie-cutter";
+import Cookies from "js-cookie";
 
 const setLanguage = (language) => {
-  language === "fr"
-    ? cookieCutter.set("lang", "fr")
-    : cookieCutter.set("lang", "en");
+  language === "fr" ? Cookies.set("lang", "fr") : Cookies.set("lang", "en");
```
var-kyle, line 15: Can you add either the `secure` or `sameSite` option to the cookie?

var-kyle (changes requested):
Looks great! I'm not running into any funny bugs anymore. Just a couple of small things before I'll approve.

var-kyle (approved):
Very nice, working as expected.

PR #440, opened 2021-12-13
Feedback accessibility

shawn320:
# Description

- Feedback no longer full width of the screen
- Give feedback popup a border
- Give the background a dark color when feedback is active

[New feedback design on Figma](https://www.figma.com/file/ke3SrZNGMP3AjwzMMuEgbo/SARAK-BDM-DECD-SC-Labs?node-id=5425%3A17621)

## Checklist

- [ ] Update CHANGELOG


krischarbonneau (changes requested):
I think this looks okay for the most part but the width of the pop up gets pretty funky once you start getting past 1200px. This is at 2560px for reference. A possible fix would be to do `md:w-4/5 xxl:w-max` which looks pretty good at higher resolutions but causes the pop up to be full screen width at 1200px. Maybe write a custom CSS class using the `@media` at-rule to change the width at higher breakpoints.

Screenshot 2021-12-13 104106

shawn320:
> I think this looks okay for the most part but the width of the pop up gets pretty funky once you start getting past 1200px. This is at 2560px for reference. A possible fix would be to do `md:w-4/5 xxl:w-max` which looks pretty good at higher resolutions but causes the pop up to be full screen width at 1200px. Maybe write a custom CSS class using the `@media` at-rule to change the width at higher breakpoints.
> 
> Screenshot 2021-12-13 104106

Thank you for pointing this out, I removed md:4/5 and used horizontal margin instead. The width on feedback is now `w-auto`, so it won't get stretched when resize screen.

krischarbonneau (dismissed):
Looks good. Great job!

krischarbonneau (approved):

PR #439, opened 2021-12-09
Signup info

krischarbonneau:
# Description

[Dev for SCL-468](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-542)
[Dev for SCL-466](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-557)

This PR adds a sign up info page which the user is shown first before the sign up form. I also removed the initial info from the sign up page (since it's all on the info page now).

I've commented out all the metadata on this page as that will be done in a separate PR based on the information provided on the Jira task.

## Acceptance Criteria

Get the acceptance criteria for the story and/or task you are working on and put it all here.

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3. Click on `Sign up` from the menu or navigate to `/signup-info`

## Help Requested

I have no idea what the french path is supposed to be so I have it as `inscription-info` right now as a placeholder.

## Checklist

- [x] Strings use placeholders for translation (No hard coded strings)
- [x] Unit tests have been added/updated
- [x] Update CHANGELOG

## Product and Sprint Backlog

[Jira](https://jira-dev.bdm-dev.dts-stn.com/secure/RapidBoard.jspa?rapidView=96&projectKey=SCL)

PR #441, opened 2021-12-13
Fixed button on smaller screens

krischarbonneau:
# Description

Quick PR to address @shawn320 's comment on branch #439 that was made after I merged. I adjusted padding on the Continue button on `signup-info` based on breakpoints so it no longer goes off screen. I also attempted to do this on the `signup` page but also found that a lot of text is overlapping <280px so that will have to be fixed in a separate PR. 

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/signup-info`
4. Adjust screen size to ensure button never overlaps or goes off screen
5. Do the same on `signup` page (going below 280px causes a lot of the other text on the page to start overlapping as mentioned above, I've only adding conditional sizing based on breakpoints for the button)

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog


shawn320 (approved):
Looks great, thank you.

PR #443, opened 2021-12-14
expand and collap state on checkbox

shawn320:
# Description

Added `aria-label` attribute on the checkbox on the Report A Problem form. The value of aria-label will either be `collapsed` or `expanded` depend on the state of the checkbox. By default, it's `collapsed`. When checkbox is checked, the value will be `expanded`

I tried to use `aria-expanded` at the beginning, but this attribute does not work on checkbox. I've tested it on different browser, but `aria-expanded` seems only work on buttons. The screen reader won't anounce it on checkbox.

## Test Instructions

Open the Report A Problem form on any pages. Test with screen reader

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

[Report A Problem Accessibility Update](https://jira-dev.bdm-dev.dts-stn.com/secure/RapidBoard.jspa?rapidView=96&projectKey=SCL&view=detail&selectedIssue=SCL-573)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # Changed
 
+- Add collape and expand state on the checkbox on the Report A Problem form
 - Add border to feedback popup, and give background a darker color
 - Lock scroll when feedback popup is active
 - Deactivate auto focus when feedback popup is active
--- a/components/atoms/CheckBox.js
+++ b/components/atoms/CheckBox.js
@@ -24,7 +24,6 @@ export function CheckBox(props) {
         name={props.name}
         value={props.value}
         type="checkbox"
-        aria-label={props.ariaLabel}
         onChange={(e) =>
           props.onChange(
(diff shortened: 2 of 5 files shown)


var-kyle (changes requested):
We want to use `aria-expanded` instead of `aria-label` in this situation. aria-label would just override the actual label and all a screen reader would say is "collapsed" or "expanded".

aria-expanded just has boolean values, so it should be `false` by default and then set to `true` when the checkbox is selected.

var-kyle:
> We want to use `aria-expanded` instead of `aria-label` in this situation. aria-label would just override the actual label and all a screen reader would say is "collapsed" or "expanded".

My mistake, `aria-label` won't override the label, I was thinking of `aria-labelledby`

shawn320:
> 

Made a small adjustment, instead of using `aria-label` which overrides the original label, I added a `sr-only` text under the label. The screen reader will anounce either `expanded` or `collapse` depends on the state change.

var-kyle (approved):
I would still prefer using `aria-expanded`, but it doesn't work with Mac's VoiceOver, so this is a good substitution. Thank you for all the testing to find that out @shawn320

PR #447, opened 2021-12-20
submit feedback bug fix

shawn320:
# Description

[Feedback submit bug](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-581)

After a successful or failed submit, if the user try to submit the feedback again, the "This field is required" error will show up even if the input still exists. The purpose of this pr is to fix this bug.

- If the feedback successfully submited, the input field should be cleared.
- If the feedback submit is failed, the input should remain, but the required error should not show up when submit again.
- When the feedback popup is closed, the input should be cleared as well. 

## Checklist

- [x] Update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Fixed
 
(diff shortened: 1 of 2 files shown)


var-kyle (approved):
Nice catch, I think these changes make a lot of sense.

PR #456, opened 2022-01-10
Update required fields

EphraimAssi:
# Description

[Add proper invalid attribute to required fields on error](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-574)

Set an area-invalid attribute to the inputs containing errors when the form is submitted with errors.

[Add proper invalid attribute to required fields on error](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-574)

Add aria-hidden the * before the label and the (required) after the label, and change the HTML for the inputs from required="required" to aria-required="true".

## Test Instructions

1. Run the code locally
2. Go to the signup form
3. Check the required fields to make sure the * before the label and the (required) after the label have an aria-hidden attribute
4. Then check to see if the inputs have the aria-required="true"
5. After that, submit the form without filling it
6. Check to see if the required fields have an area-invalid attribute set to the inputs with errors

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog


krischarbonneau (changes requested):
Hey Ephraim, this is good however after a meeting with Sara and Celeste there is a slight design change. 

The proposed design change is like the below image from [this page](https://www.nngroup.com/articles/required-fields/).
Screenshot 2022-01-11 120225

The idea here is that we don't need both the asterisk and the `(required)` in the label. Normally we would just remove the `(required)` and put an `aria-required=true` however [this page](https://dequeuniversity.com/tips/identify-required-elements) mentions that some older screen readers that don't support ARIA may not announce the word required unless it is in the label. While this is becoming less and less of an issue as time goes on (see: [https://www.powermapper.com/tests/screen-readers/aria/](https://www.powermapper.com/tests/screen-readers/aria/)), perhaps we visually hide the word `(required)` using `sr-only` for the time being rather than solely rely on aria-required.

components/atoms/SelectField.js
```diff
@@ -37,7 +37,7 @@ export function SelectField(props) {
         ) : undefined}{" "}
         {props.label}{" "}
         {props.required ? (
-          <b className="text-error-border-red" aria-hidden="true">
+          <b className="text-error-border-red sr-only" aria-hidden="true">
```
krischarbonneau, line 40: Having both `sr-only` and `aria-hidden="true"` here makes this invisible to both screen readers and regular users. I've somewhat flip flopped on this but after some thought, I think given that we have `aria-required` on the input itself and seeing that most popular screen readers have support for `aria-required`, I'd say we just get rid of this line (and any red `required` on the other components).

krischarbonneau (changes requested):
The design change looks good, only one comment regarding the `(required)` on labels.

pages/signup.js
```diff
@@ -613,8 +619,7 @@ export default function Signup(props) {
                   className="block leading-tight text-sm font-body mb-5 lg:text-p font-bold"
                   aria-required="true"
```
krischarbonneau, line 620: For whatever reason this won't announce required now on the grouping, but it works if you set `aria-required="true"` on the `fieldset` instead

krischarbonneau (changes requested):
Almost there. Just a couple minor changes to `signup.js` and I'll approve. In addition to the changes requested above, the "agree to conditions" checkbox should have the required attribute (at line 1126) so that `aria-required` is set on it.

pages/signup.js
EphraimAssi: `aria-required="true"` on the `fieldset` creates an a11y cypress error, but it works if we put required directly on the radio inputs.

krischarbonneau: This isn't what we want either though since screen readers would announce required on both radio buttons, which would be confusing for screen reader users. This was pointed out to us as an example when we took a11y training and was originally addressed in #457 but that solution no longer seems to work in this PR.

krischarbonneau: I've tried a few things to get this working to no avail. For whatever reason NVDA will no longer register that `aria-required` is present on the `legend`. A workaround for now would be to do something like below and create a bug issue:

```html                
<legend className="block leading-tight text-sm font-body mb-5 lg:text-p font-bold">
  <b className="text-error-border-red" aria-hidden="true">
     *
  </b>{" "}
  {t("formLang")}
  <span className="sr-only">required</span>
</legend>
```

krischarbonneau (approved):
Looks good now, great job!

shawn320 (approved):
Nice work, thank you.

PR #457, opened 2022-01-11
Added aria-required to language question

krischarbonneau:
# Description

[Use aria-required on the legend instead of the radio buttons for the language preference question](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-575)

The purpose of this PR is to fix an issue where screen readers announce "required" on both radio button options when only one is required to be selected, which could lead to confusion. By removing `required` from both radio buttons and adding `aria-required="true"` to the legend, the screen reader will announce that the grouping is required rather than each individual radio button.

## Acceptance Criteria

Screen reader only announces required once on the grouping

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/signup`
4. Boot up NVDA
5. Tab to first language radio button
6. Confirm that NVDA announces `required grouping` and doesn't announce required when alternating between radio options

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Removed all the initial text on signup page
 - Updated metadata on every page according to the master content inventory
 - Updated project to use Tailwind V3
+- Updated the language question on the signup page by putting `aria-required` on the legend and removing `required` from each radio button so screen readers only announce required once for the grouping
 
 ## Fixed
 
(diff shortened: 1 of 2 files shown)


var-kyle (approved):
Looks great, glad to remove the duplicate announcing of "required" for screen readers. I do wonder if it's necessary to turn the radio button text red when there is an error though.

krischarbonneau:
> Looks great, glad to remove the duplicate announcing of "required" for screen readers. I do wonder if it's necessary to turn the radio button text red when there is an error though.

I think it's at least consistent with the text fields and dropdowns having a red border but we could always look at changing it.

PR #458, opened 2022-01-11
Added autocomplete prop to TextField

krischarbonneau:
# Description

[Add autocomplete to appropriate inputs](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-571)

Small PR that adds the `autoComplete` prop to the TextField component. By default it is set to "off" (which may not do a whole lot depending on which browser you use since Chromium browsers tend to ignore this). At this point in time this really doesn't do a whole lot since we really only have the email fields on the signup form that can take advantage of this but this will let us tailor autocomplete to our needs if we add new fields in the future.

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3. If your browser does not already have saved information, fill out the form once with information and press submit (else you should see your saved e-mail pop up when clicking on the email field)
4. Click on the email field(s) and ensure your information pops up
5. Try changing the value for `autoComplete` from `email` to `country` and click on the field to see the autocomplete suggestion change

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Added metadata for `/500`, `/error`, `/404`, `/notsupported`, `/projects/virtual-assistant`, `/projects/digital-centre`, `/signup-info`, and the splash page
 - Added an optional `Are you a public servant` question to sign-up form
 - Added an `Confirm email` field to sign-up form
+- Added `autoComplete` prop type to TextField component
 
 # Changed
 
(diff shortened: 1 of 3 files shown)


var-kyle (commented):
I am curious what you're thoughts are of just not adding the autocomplete when it doesn't exist vs setting it to "off" by default.

My opinion is that it shouldn't be added if someone doesn't want it there. The way you have it set up, it can't be removed from any text field.

krischarbonneau:
> I am curious what you're thoughts are of just not adding the autocomplete when it doesn't exist vs setting it to "off" by default.
> 
> My opinion is that it shouldn't be added if someone doesn't want it there. The way you have it set up, it can't be removed from any text field.

Yeah, this makes more sense. No real reason to have it as off by default vs just not setting it.

var-kyle (approved):
I was worried that if the autocomplete prop was not included it would result in `autocomplete=undefined`, but that doesn't seem to be the case.

Awesome work, thank you!

PR #459, opened 2022-01-13
strapi service class

shawn320:
# Description

[Create a Strapi Service](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-589)

Created a class that uses Singleton pattern to manage data from Strapi.

The StrapiService class will be instantiated during runtime, returns the project data that fetched from Strapi. The instance of the class will be created if it does not exist, otherwise, return the existing one. The class uses cache busting strategy to allow browser to get the most recently updated data from Strapi. The get method returns the stored data, if the data does not exist, it will be fetched from Strapi.

Note that I did not create this class, the Digital Centre project already has this that works with AEM, I believe it was created by @zeddotes, so the class I created is pretty similar to the one that the Digital Centre project has. 

## Test Instructions

Strapi was down at the time when I was testing this. So I used the mocked api, you just need to assign `enabled` to the `NEXT_PUBLIC_API_MOCKING` variable. You also need to add a new variable in `.env` named `NEXT_PUBLIC_BUILD_DATE`, check the `.env.example` file.

## Help Requested

- Things that you (the submitter)...
- want reviewers to pay very close...
- attention to when they review this PR.

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog


zeddotes (approved):
This is great stuff @shawn320. Thanks

PR #462, opened 2022-01-20
Re-send bug fix

shawn320:
# Description

[Re-send email for unsubscribe](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-600)

Fixed the issue where re-send email for unsubscribe will get a signup confirmation email.

Note: I will open a new pr to update design once it gets approved by Celeste.

## Acceptance Criteria
When request to re-send email for unsubscribe, should receive an unsubscribe email instead of confirmation email.


pages/thankyou.js
```diff
@@ -26,13 +26,24 @@ export default function Confirmation(props) {
...
+        },
+        body: JSON.stringify({ email: email }),
+      });
+    } else {
+      response = await fetch("/api/thank-you", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({ email: email, ref: referrer }),
```
var-kyle, line 44: You can remove the ref parameter in the body, since the api does nothing with it.

var-kyle (changes requested):
I kind of thought since you created the /thank-you api route, the logic was going to be handled there, but it works here as well. Just a small thing to clean up and should be good to merge.

pages/thankyou.js
shawn320: > I kind of thought since you created the /thank-you api route, the logic was going to be handled there, but it works here as well. Just a small thing to clean up and should be good to merge.

Nice catch, I was going to handle the logic in the /thank-you api, then I notice /unsubscribe api already handles it, so I decided to just use that.

var-kyle (approved):
Awesome, looks good! Thank you

PR #466, opened 2022-01-27
Fix focus on feedback widget

EphraimAssi:
# Description

[Focus Indicator Disappears While Tabbing](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-606)

Fixed the focus indicator on the feedback widget on the close button and the email "experience@servicecanada.gc.ca"

## Test Instructions

1. Run the code locally
2. Check the feedback widget to see if the close button and the email "experience@servicecanada.gc.ca" focus on tab

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog


components/molecules/FeedbackWidget.js
```diff
@@ -188,32 +190,34 @@ export const FeedbackWidget = ({
...
+                    <ActionButton
+                      id="link-mail"
+                      dataCy="link-mail"
+                      dataTestId="link-mail"
                       href={`mailto:${email}`}
-                      aria-label="Service Canada email"
-                    >
-                      {email}
-                    </a>
+                      ariaLabel="Service Canada email"
```
var-kyle, line 217: ActionButton renders as an `<a>` tag here, with text inside of it so there shouldn't be an `aria-label`.

components/molecules/FeedbackWidget.js
```diff
@@ -188,32 +190,34 @@ export const FeedbackWidget = ({
...
+                <ActionButton
                   id="feedbackClose"
+                  dataCy="closeButton"
+                  dataTestId="closeButton"
+                  ariaLabel="Close the expanded feedback section"
+                  custom="flex float-right pt-2 font-body text-white flex mt-2.5 lg:mt-0 outline-none focus:outline-white-solid items-center"
+                  imageSource="/close-x.svg"
+                  imageAlt="Close button"
+                  imageSpanClass="text-xs leading-4 lg:text-sm underline ml-2 lg:leading-10"
+                  imageSpanText={t("close")}
```
var-kyle, line 202: `aria-label` should only be used on elements that don't have text, or don't have a label associated with an image. In this case, there is a span inside the button with text, which accurately describes what the button does, so `aria-label` shouldn't be necessary here.

var-kyle (changes requested):
I like that you're making use of the ActionButton here, rather than `<button>` and `<a>` tags, it makes sense to use the components we have. I am a little concerned that the ActionButton is going to become to big, or hard to use since it has a lot of logic in it. But that's for some other time. The only thing I noticed was the use of `aria-label` that I think it incorrect.

krischarbonneau (changes requested):
Focus seems to work well now but NVDA announces "Undefined button" on every `ActionButton` so it the `imageSpanText` prop doesn't seem to be working

EphraimAssi:
I'm not sure why it's acting like that, but I put back the Aria-Labels

krischarbonneau (changes requested):
The submit button still says `undefined button`, otherwise I think this is good

components/molecules/FeedbackWidget.js
```diff
@@ -275,6 +275,7 @@ export const FeedbackWidget = ({
                   />
                   <ActionButton
                     id="feedback-submit"
+                    ariaLabel="Submit feedback button"
```
krischarbonneau, line 294: Should just be `Submit feedback`, otherwise screen readers will announce `Submit feedback button button`

components/molecules/FeedbackWidget.js
```diff
@@ -195,40 +193,43 @@ export const FeedbackWidget = ({
                 ""
               )}
               <div className="layout-container text-gray-dark-100 pb-4">
-                <button
+                <ActionButton
                   id="feedbackClose"
+                  ariaLabel="Close the expanded feedback section"
+                  dataCy="closeButton"
+                  dataTestId="closeButton"
+                  custom="flex float-right pt-4 font-body text-gray-dark-100 flex mt-2.5 lg:mt-0 outline-none focus:outline-white-solid items-center"
```
krischarbonneau, line 201: The focus outline on the close button doesn't look that great. A work around here would be to get rid of the `pt-4` from the custom attribute and instead put this ActionButton inside a div with top padding. 

ie.
```JSX
              <div className="layout-container text-gray-dark-100 pb-4">
                <div className="pt-2">
                <ActionButton
                  id="feedbackClose"
                  ariaLabel="Close the expanded feedback section"
                  dataCy="closeButton"
                  dataTestId="closeButton"
                  custom="flex float-right font-body text-gray-dark-100 flex mt-2.5 lg:mt-0 outline-none focus:outline-white-solid items-center"
                  imageSource="/close-x.svg"
                  imageAlt="Close button"
                  imageSpanClass="text-xs leading-4 lg:text-sm underline ml-2 lg:leading-10"
                  imageSpanText={t("close")}
                  onClick={() => {
                    toggleForm();
                    setCount(2000);
                  }}
                />
                </div>
```


Screenshot 2022-02-02 151703

krischarbonneau (commented):
A couple more changes and this is good

PR #469, opened 2022-02-01
Changed feedback widget color scheme

krischarbonneau:
# Description

[Text elements must have sufficient color contrast against the background](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-605)

Our a11y audit came back with a suggestion to increase the contrast ratio of the `(required)` label on our feedback widget so it complies with AAA (7:1 contrast ratio, or 4.5:1 if the text is considered "Large text", which is 18px and bold). Rather than changing just the required label, we're using this task as an opportunity to improve the overall readability of the feedback widget by changing the color scheme to a more common design pattern (black text on white background).

In addition to the above, this PR also adds a remaining character counter which changes whenever the `textarea` element receives any input. I've also removed the `(required)` label and added a red star in it's place to stay consistent with the signup page's design.  

I've added a CSS class to change the color of the close ( X ) SVG when successfully submitting feedback, but since the plan is to eventually do away with the success banner and change the content within the window, this is only temporary.

## Test Instructions

1. Pull in branch
2. Type `npm run dev`
3. Navigate to `/projects/digital-centre`
4. Open feedback widget
5. Start typing, remove text, paste text into box to ensure character counter works as expected

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Added an `Confirm email` field to sign-up form
 - Added a wrapper class that uses singleton pattern to manage Strapi data
 - Added `autoComplete` prop type to TextField component
+- Added remaining character counter to feedback widget
 
 # Changed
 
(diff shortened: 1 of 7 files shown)


zeddotes (approved):
Awesome work, especially on this PR's description.

PR #492, opened 2022-03-24
Remove srSpeak function

krischarbonneau:
# Description

[SCL-640 - Clean up security alerts from CodeQL](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-640)

This PR addresses the CodeQL alert https://github.com/DTS-STN/Alpha-Site/security/code-scanning/1

We previously had a function that would make the screen reader announce the error label when a user hits the submit button in our feedback widget without entering feedback. This was accomplished by accessing the element in the DOM which isn't ideal. This PR does away with the function and instead just adds `aria-live = "polite"` on the entire form instead, which effectively does the same thing but with a lot less code.

The other alerts that were present were dismissed for the reasons below:

https://github.com/DTS-STN/Alpha-Site/security/code-scanning/3 - This alert was dismissed as a false positive since it thinks we're hard coding an API key when in reality we're supplying a default dummy value (as required by GCNotify). There is no security concern here.

https://github.com/DTS-STN/Alpha-Site/security/code-scanning/2 - Dismissed because this alert is not relevant. This is an icomoon demo file.

## Test Instructions

1. Pull in branch
2. Type `yarn dev`
3. Navigate to `/projects/digital-centre`
4. Open NVDA (or whatever screen reader you're using)
5. Open feedback window
6. Click Submit; The screen reader should announce the submit button's text and then the error

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Added the `stripFeedback` helper function which strips PII from both the feedback and report a problem forms
 - Added error label in report a problem widget
 
-# Changed
+## Changed
 
 - Add collape and expand state on the checkbox on the Report A Problem form
 - Add border to feedback popup, and give background a darker color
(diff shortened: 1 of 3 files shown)


zeddotes (approved):
Looks good!

PR #513, opened 2022-05-28
[Snyk] Upgrade @apollo/client from 3.6.2 to 3.6.3

DaceyTom2:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade @apollo/client from 3.6.2 to 3.6.3.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **1 version** ahead of your current version.
- The recommended version was released **22 days ago**, on 2022-05-05.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>@apollo/client</b></summary>
    <ul>
      <li>
        <b>3.6.3</b> - <a href="https://snyk.io/redirect/github/apollographql/apollo-client/releases/tag/v3.6.3">2022-05-05</a></br><h2>Bug Fixes (from <a href="https://snyk.io/redirect/github/apollographql/apollo-client/blob/main/CHANGELOG.md"><code>CHANGELOG.md</code></a>)</h2>
<ul>
<li>
<p>Simplify <code>useQuery(query, { defaultOptions })</code> default options processing in order to fix bug where <code>skip: true</code> queries failed to execute upon switching to <code>skip: false</code>.<br>
<a href="https://snyk.io/redirect/github/benjamn">@ benjamn</a> in <a href="https://snyk.io/redirect/github/apollographql/apollo-client/pull/9665" data-hovercard-type="pull_request" data-hovercard-url="/apollographql/apollo-client/pull/9665/hovercard">#9665</a></p>
</li>
<li>
<p>Add tests of skipping/unskipping and <code>useLazyQuery</code> with <code>defaultOptions</code>, and fix a bug causing duplicate requests.<br>
<a href="https://snyk.io/redirect/github/benjamn">@ benjamn</a> in <a href="https://snyk.io/redirect/github/apollographql/apollo-client/pull/9666" data-hovercard-type="pull_request" data-hovercard-url="/apollographql/apollo-client/pull/9666/hovercard">#9666</a></p>
</li>
<li>
<p>Update <code>ts-invariant</code> to version 0.10.2 to fix source map warnings.<br>
[...]

PR #514, opened 2022-06-06
Refactor notsupported page to use data from AEM

krischarbonneau:
# Description

[SCL-731 - Refactor /notsupported page to use queried data](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-731)

The purpose of this PR is to refactor the notsupported page so it uses the content and assets from AEM rather than being hardcoded. I added a mock data store for testing purposes which will grow in size as we continue to update our site to use data from AEM, though we should really look into mapping our data/creating a service class eventually. I've also updated apollo client to 3.6.6 as per #513. 

I didn't touch any of the metadata since since a meeting regarding that still needs to happen with Isabelle, so eventually our `getStaticProps` will be rid of the `serverSideTranslations` lines as they won't be necessary anymore.

It appears as if our content security policy was implemented in both our custom document file and in our next config. This is honestly pretty messy and we should look to move all of our security headers into our next config instead.

## Test Instructions

1. Pull in branch
2. Type `yarn install --frozen-lockfile`
3. Type `yarn dev`
4. Navigate to `/notsupported`
5. Ensure the page is rendered properly with the correct content

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Update signup-info page to read content from cms
 - Removed `experience` email from privacy page, Report a Problem and Feedback components
 - Updated to Next.js `v12.1.6`
+- Updated `/notsupported` to use data from AEM
 
 ## Fixed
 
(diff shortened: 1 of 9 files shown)


shawn320 (approved):
Tested, content displays properly, nice work!

PR #534, opened 2022-06-28
Projects redesign

shawn320:
# Description

[Projects page redesign](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-671)

Update the content and layout for the projects page to match the re-design on [Figma](https://www.figma.com/file/ke3SrZNGMP3AjwzMMuEgbo/SARAK-BDM-DECD-SC-Labs?node-id=9561%3A25275). Page content now using data from AEM. My previous pr for the card component had some styling issues, fixed in this pr. French content are not available at the moment, will update again once the French content get approved.

Screen Shot 2022-06-28 at 1 10 08 PM

Note: Regarding the disclaimer banner on the top, it has it's own ticket on jira. I'm not sure if it is ready for the dev, will ask Celeste. 

## Test Instructions

1. yarn dev
2. go to projects page
3. make sure it matches the design on Figma

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/pages/projects.js
+++ b/pages/projects.js
@@ -219,7 +219,7 @@ export default function Projects(props) {
 }
 
 export const getStaticProps = async ({ locale }) => {
-  // get projects data from AEM
+  // get page data and experiments data from AEM
   const res1 = await queryGraphQL(getAllProjects).then((result) => {
     return result;
   });
--- a/pages/projects.js
+++ b/pages/projects.js
@@ -164,7 +164,7 @@ export default function Projects(props) {
               </p>
             </span>
             <span
-              className="relative flex w-full lg:ml-8"
+              className="relative flex w-full mt-4 lg:ml-8"
               style={{ height: "274px", width: "453px", minWidth: "453px" }}
             >
               <Image
--- a/pages/projects.js
+++ b/pages/projects.js
@@ -154,15 +154,21 @@ export default function Projects(props) {
         </Head>
         <section className="layout-container mb-10">
           <h1 id="pageMainTitle" tabIndex="-1" className="flex-wrap mb-2">
-            {pageData.scTitleEn}
+            {props.locale === "en" ? pageData.scTitleEn : pageData.scTitleFr}
           </h1>
           <div className="lg:flex">
             <span className="w-full py-4">
               <p>
-                {pageData.scFragments[0].scContentEn.json[0].content[0].value}
+                {props.locale === "en"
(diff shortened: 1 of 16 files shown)


pages/projects.js
```diff
@@ -173,36 +204,28 @@ export default function Projects(props) {
...
+                  title={experiment.title}
+                  tag={experiment.tag}
+                  tagLabel={t(experiment.tag)}
+                  description={experiment.description}
+                  href={experiment.href}
+                  dataTestId={`${experiment.id}`}
+                  dataCy={`${experiment.id}`}
+                  isExperiment
+                  imgSrc="/placeholder.png"
+                  imgAlt="placeholder"
```
krischarbonneau, line 223:
```suggestion
                  imgAlt="placeholder"
                  priority
```

krischarbonneau (commented):
This looks great, though there's a few console warnings/errors we should clean up:

Some image resources are getting blocked on page load. Change the CSP in `next.config.js` to the below:
```
    key: "Content-Security-Policy",
    value: `default-src 'self' dts-stn.com *.dts-stn.com *.adobe.com *.omniture.com *.2o7.net; frame-ancestors 'self'; base-uri 'self'; form-action 'self'; connect-src 'self' *.demdex.net *.omtrdc.net cm.everesttech.net; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com data:; img-src 'self' data: *.omtrdc.net *.demdex.net cm.everesttech.net assets.adobedtm.com https://www.canada.ca; font-src 'self' https://fonts.googleapis.com https://fonts.gstatic.com; frame-src 'self' *.demdex.net; script-src 'self' 'unsafe-inline' ${
      process.env.CI === "true"
        ? "'unsafe-eval'"
        : process.env.NODE_ENV === "development"
        ? "'unsafe-eval'"
        : ""
    } https://ajax.googleapis.com https://assets.adobedtm.com;`,
```

The second thing that pops up is that the placeholder image is detected as an LCP (Largest Contentful paint) element and as such should be given the priority flag so it is not lazy loaded.

Since the priority flag won't always be needed, I recommend changing the Image tag within `Card.js` to the below:
```
<Image
          src={props.imgSrc}
          alt={props.imgAlt}
          layout="fill"
          objectFit="cover"
          priority={props.priority}
        />
```
and specifying priority on the Card element in `projects.js` (see other comment)

PR #535, opened 2022-06-28
Updated homepage as per new designs

krischarbonneau:
# Description

[SCL-667 - Dev for homepage](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-667)

The purpose of this PR is to update the home page to match the new Figma designs. We do not have complete translations yet so switching to French will show either no content or English content with `(FR)` at the end. Metadata terms are also not completed on the AEM side yet so those are also not in this PR. I've also removed the `Menu` portion from our `Layout` component as per Celeste and new designs and added more padding to the breadcrumbs to better match designs.

I've chosen to hide the green image on screens widths below 992px for the time being but this may change depending on what Celeste wants.

## Test Instructions

1. Pull in branch
2. Type `yarn dev`
3. Navigate to the home page and ensure it matches the new designs

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Updated `/signup` page to use get data from AEM
 - Updated `/notsupported`, `/404`, `/500`, and `/error` to use data from AEM
 - Updated Experiment component to a generic card component
+- Updated home page to match new Figma designs
 
 ## Fixed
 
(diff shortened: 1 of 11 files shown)


krischarbonneau:
For whatever reason, the cypress tests on the home page that look for the buttons to the about, projects, and sign-up pages are timing out and causing the the cypress step to fail. Not sure why this is happening since the identifier is correct in both the test file and on the home page, so I've commented the problem tests out for now to get this PR moving and will look into these tests separately since they aren't really all that relevant. Issue is perhaps related to AEM.

shawn320 (dismissed):
Looks good to me, nice work! The buttons on the French page has not text, but since we don't have the French content yet, it's okay like this.

shawn320 (approved):

PR #536, opened 2022-06-28
Added overwrite functionality to date modified component

krischarbonneau:
# Description

[SCL-747 - Investigate "date modified" component](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-747)

The purpose of this PR is to update the `DateModified` component to accept a manual date override, rather than solely rely on the build date from Azure. I've added a default prop that takes the Azure build time and added a new prop to the `Layout` component that accepts a date string in the format `yyyy-mm-dd`. Since the format for a date string from AEM follows the `yyyy-mm-dd` format and azure follows the `yyyyMMdd` format, I've introduced a regex pattern that is matched against the date prop which checks to see if the string contains dashes. If not, then dashes are added so it follows the `yyyy-mm-dd` format.

Since not all pages have been updated yet, I've only updated the home page to use the date override as an example.

## Test Instructions

1. Pull in branch
2. Type `yarn dev`
3. Navigate to the home page
4. On line 28 of `home.js`, change the value for `dateModifiedOverride` to "2022-06-25" and refresh the page. Ensure this shows up properly
5. Now update the value to "20220626", the regex pattern should see that there are no hyphens in the string and add them so it appears as `2022-06-26` on the screen

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Updated `/notsupported`, `/404`, `/500`, and `/error` to use data from AEM
 - Updated Experiment component to a generic card component
 - Updated home page to match new Figma designs
+- Updated DateModified component to accept a manual date (for use with AEM)
 
 ## Fixed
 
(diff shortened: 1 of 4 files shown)


shawn320 (approved):
Nice work! It's unfortunate that we can't grab the built in date modified on AEM, but this is good for now. It also makes much more sense that each page has it's own date modified.

PR #555, opened 2022-07-25
Clean up environment variables

will0684:
# Description

[Clean up environment variables - SCL-598](https://jira-dev.bdm-dev.dts-stn.com/browse/SCL-598)

Removed unused vars
Removed build args from docker where possible
Removed references to `process.env` in pages where possible and replaced them with calls to `props`
Removed NEXT_PUBLIC_ from env vars where it was not needed

## Acceptance Criteria

- Variables that do not need to be public should have NEXT_PUBLIC_ removed.
- Variables that do not need to be consumed during build time should be removed from docker build args
- Variables that can be passed as props rather than referenced directly should be passed as props
- Variables in dev and prod pipeline should be modified to correspond with these changes

## Test Instructions

1. Add required values as default arguments to the Dockerfile i.e. `ARG AEM_GRAPHQL_ENDPOINT=https://www.someUrl.ca/endpoint.json`
2. `docker build .`
3. `docker run -p 3000:3000 <imageID>`
4. See that values were populated

## Help Requested

- Please share anything else you think could be done to further clean up the Dockerfile and build arguments

## Product and Sprint Backlog


pages/projects.js
```diff
@@ -43,6 +43,7 @@ export default function Projects(props) {
   };
 
   useEffect(() => {
+    console.log(props.adobeAnalyticsUrl);
```
krischarbonneau, line 46: Let's remove this

will0684: Oops yeah thanks

krischarbonneau (approved):
Great work!

PR #556, opened 2022-07-26
remove test site banner

shawn320:
# Description

A small pr to remove the TEST SITE banner on the projects page

## Checklist

- [x] Update CHANGELOG


krischarbonneau (approved):
This looks fine but are we removing this component entirely? The banner is not being used on any pages now and the "alert" that is going on the projects page will be halfway down the page.

shawn320:
> This looks fine but are we removing this component entirely? The banner is not being used on any pages now and the "alert" that is going on the projects page will be halfway down the page.

I didn't remove it because I was thinking we could update this component in the future for the "alert", looking at the design, it has some similarity with the test banner.

PR #561, opened 2022-08-04
Updated sign up banner using FontAwesome icon

krischarbonneau:
# Description

[64042 - DEV: Update "sign up" banner (front end design)](https://dev.azure.com/VP-BD/DECD/_sprints/taskboard/Service%20Canada%20Labs/DECD/Service%20Canada%20Labs/PlanningSprint-1?workitem=64042)

In this PR, I've added Font Awesome to our app in order for the sign up banner to match the design in Figma. This will also be used for the contextual alert on the `/projects` page. 

Since the designs on Figma are purely for Desktop, I've gone ahead and scaled the text down for mobile as I saw fit, input here would be appreciated.

Since we currently don't know what exactly is to happen with the feedback component, I haven't touched the related code in the `CallToAction` component (aside from padding and text size just so it doesn't look like garbage). If it's decided that we will be keeping it/repurposing it, we can then revisit this and adjust the `CallToAction` component. Currently though, there are no designs. If you're curious though, you can check out `/projects/digital-centre` to see what it currently looks like with the new design.

## Test Instructions

1. Pull in branch
2. Type `yarn dev`
3. Navigate to `/projects`
4. Scroll to the bottom of the page to see the new sign up banner
5. Open developer tools and play around with different resolutions to make sure it shows up properly/scales well

## Notes

- A fragment holding the sign up banner text will eventually need to be created within AEM if it hasn't already been created (it may already be created using the `feature` model and if that's the case we would just have to add that fragment to our page fragment)

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

### DIFF
--- a/components/molecules/CallToAction.js
+++ b/components/molecules/CallToAction.js
@@ -2,6 +2,8 @@ import PropTypes from "prop-types";
 import { ActionButton } from "../atoms/ActionButton";
 import Link from "next/link";
 import { useTranslation } from "next-i18next";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faUser } from "@fortawesome/free-regular-svg-icons";
 
 /**
  * A section that will have a title, small description, and a link to some action we want to user to make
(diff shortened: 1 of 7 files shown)


components/molecules/CallToAction.js
```diff
@@ -2,65 +2,81 @@ import PropTypes from "prop-types";
...
-              <p className="whitespace-pre-line">{props.description}</p>
-            ) : (
-              <div
-                className="whitespace-pre-line"
-                dangerouslySetInnerHTML={{ __html: props.html }}
+    <div className="bg-[#ebf2fc] text-[#333]">
+      <div className="layout-container pb-10 pt-10 text-sm md:text-base">
+        <div>
+          <div className="flex flex-col">
+            <div className="flex -my-2">
```
will0684, line 18: This y margin isn't being applied due to the preceding "-"

will0684 (changes requested):
Besides the one tiny thing it looks awesome, its responsive and working. Great work.

components/molecules/CallToAction.js
krischarbonneau: This is intended, see https://tailwindcss.com/docs/margin#using-negative-values

will0684: Oh well then TIL!

will0684 (approved):
Nice work 👍

PR #562, opened 2022-08-05
Updated card component and GraphQL page queries

krischarbonneau:
# Description

This PR addresses a concern that Celeste had with the images within our Card component (specifically the blue paper plane one). This issue was also present on the project page and has to do with how we set up our pages to use `Next/Image`. Previously, we were using `layout="fill"` and `objectFit="cover"` to maintain aspect ratio when scaling through different viewport sizes. This worked okay until you get down to a mobile resolution and then the sides of the image get cut off.

The solution here was to switch to `layout="responsive"` and provide the length and width of the image straight from AEM data (since Next/Image requires you to explicitly provide this if you are not using `layout="fill"`) for proper scaling.

That said, I've made a couple changes to the queries themselves, notably adding `height` and `width` to the image fields and changing `_path` to `_publishUrl` which is cleaner. Note that once we remove the placeholder images from the project cards and supply them with images from AEM, we will need to also grab the height and width for those images from AEM.

One thing of note is that while AEM supplies the height and width of images when querying using `... on ImageRef`, there is no equivalent for `... on DocumentRef`, so this way of obtaining the dimensions of the image ONLY works if we are not using an SVG.

## Test Instructions

1. Pull in branch
2. Type `yarn dev`
3. Navigate to both `/home` and `/projects`
4. Open developer tools to view the page in different viewports or resize your window to ensure the images resize properly and aren't being cut off

## Checklist

- [x] Update CHANGELOG

### DIFF
--- a/pages/projects.js
+++ b/pages/projects.js
@@ -18,6 +18,7 @@ export default function Projects(props) {
   const [filteredExperiments, setFilteredExperiments] = useState(
     props.experimentData.items
   );
+  console.log(pageData.scFragments[3].scImageEn._publishUrl);
 
   // get the filters from the data
   const filters = props.filters.map((value) => {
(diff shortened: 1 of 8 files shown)


pages/projects.js
```diff
@@ -18,6 +18,7 @@ export default function Projects(props) {
   const [filteredExperiments, setFilteredExperiments] = useState(
     props.experimentData.items
   );
+  console.log(pageData.scFragments[3].scImageEn._publishUrl);
```
will0684, line 21: Did we want this console.log?

will0684 (commented):
Just one small comment but looks and works good!

pages/projects.js
krischarbonneau: Good eye, I'll remove this!

will0684 (approved):
Nice work!

PR #569, opened 2022-08-16
Social media image

shawn320:
# Description

The purpose of this pr is to update the social media image for the splash page. The new image should be the same as the background image for splash page.  I'm also trying to fix the broken image issue on the Having a child page. The first image returns 403, happens in prod only, so I'm not able to test it locally. I set high priority on the images in the Having a child page, so that lazy loading is disabled, hopefully this could fix something. 

## Checklist

- [ ] Update CHANGELOG


public/locales/en/common.json
```diff
@@ -307,7 +307,7 @@
   "before": "Before",
   "after": "After",
   "contactUsResponse": "You will receive a response within 3-5 business days from Monday to Friday, except holidays.",
-  "metaImage": "/public/SCLabs-meta-image-EN.jpg",
+  "metaImage": "/public/sclabs-meta-image.png",
```
krischarbonneau, line 310: I think this is okay for now, but realistically this image should come from AEM

krischarbonneau (approved):
Looks good, fingers crossed the issue on the Having a Child page is solved with this

PR #570, opened 2022-08-16
Image return 403 error

shawn320:
# Description

Opened this pr for testing, I temporarily use html img instead of Next image for the first image in the Having a child page, so that I can see if the image still return 403 in prod.


krischarbonneau (approved):
Would be strange if it was purely related to Next/Image, but let's give this a go.

PR #577, opened 2022-09-02
Adding en to URL path by default

krischarbonneau:
# Description

[71053 - Include EN in english language URL path](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities/?workitem=71053)

After some discussion I've decided to remove redirects based on language stored in cookies. Next.js' current official workaround/implementation for having `/en` in the URL is to create a dummy language (in this case `default`) and then use a middleware that is called on every route using the `default` locale and redirect to the `en` page. As an example, our home page using the default locale would just be `/home` but the middleware takes this request and returns the new URL `/en/home`.
This works fantastic, until it doesn't. Next.js has an execution order unfortunately which runs any redirects()  in our next.config.js first, and then any redirects() in our middleware. This causes some really wonky behavior when initially switching to an English page from a French page.
Now you may ask: "Well why can't we just do all the redirects inside the middleware instead of in our next.config.js?"
We can! In fact this method works perfectly....in a development environment only.
The below code block is a rough example of how we could accomplish this on the home page:
 ```
if (req.nextUrl.pathname === "/home" && req.cookies.get("lang") === "fr") {
    return NextResponse.redirect(new URL(`/fr/accueil`, req.url));
  } else if (
    req.nextUrl.pathname === "/accueil" &&
    req.cookies.get("lang") === "en"
  ) {
    return NextResponse.redirect(new URL(`/en/home`, req.url));
  }
```
where `req.nextUrl.pathname` grabs the path name of the URL (ie. `/home`), determines if it should be on the English or French page and then redirects to the proper page if necessary depending on the language currently in the cookie.
The problem with this solution is that it doesn't work in a production environment, like at all. For some reason, in a production environment only, the value for `req.nextUrl.pathname` updates every time you hover over a link, which messes up the logic completely.
All that said, [Canada.ca](http://canada.ca/) doesn't store any language cookie when selecting a language on their splash page, so I'll be removing cookies for now in favor of bilingual path names and to be consistent.

This PR also updates Next.js to 12.2.5 (this was as a result of testing to make sure errors weren't caused by an outdated version of Next, but I've left it in this PR since its good to keep it up to date)

## Acceptance Criteria

An English URL should have `/en` in it's path 

## Test Instructions

1. Pull in branch
2. Type `yarn install --frozen-lockfile`
3. Type `yarn build` and `yarn start`
4. Start by selecting a language and then navigate throughout the site, ensuring that English URLs contain `/en` and French URLs contain `/fr`

## Checklist

- [x] Update CHANGELOG

## Product and Sprint Backlog

[ADO](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities)


will0684 (approved):
Tested and working perfect, great work 👍

PR #582, opened 2022-09-14
AEMService

zeddotes:
- updated revalidation logic to get a query param for routes (comma-separated); will run revalidation on the supplied URLs rather than trying to maintain a changing URL pattern between environments (next.js issue)
- storing graphql cache in its own folder, added to .gitignore
- updated pages with the query using the new AEM service
- updated tests


krischarbonneau (approved):
Works well! Great job on this.

PR #613, opened 2022-10-20
Created new model for Update/Blog pages and modified query for VA Update pages

will0684:
# Description

[Dynamic page model for blog](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities/?workitem=63969)

Created a new model for blog/update pages called "SCLabsBlog-v1" which includes two new single line string fields. Also created a new query for Virtual Assistant update pages which will only return pages related to VA from all the blog content fragments. 

The new string fields are as follows:

`scProject` is where you enter the string identifying the project with which this update page should be associated
`scProjectPhase` is where you enter the string identifying the phase that the project is in

I have also added these fields to the Project Update cards content fragment model and their query so that we can eventually filter them on the frontend similar to how we filter project cards based on project phase.

## Acceptance Criteria

The Virtual Assistant update page should work bilingually with the new model and query.

## Test Instructions

1. yarn dev
2. navigate to Virtual Assistant update page through the card
3. page should be seen working in both languages

## Product and Sprint Backlog

### DIFF
--- a/graphql/queries/projectUpdatesQuery.graphql
+++ b/graphql/queries/projectUpdatesQuery.graphql
@@ -6,6 +6,8 @@ query getProjectUpdates {
       scTitleEn
       scTitleFr
       datePosted
+      scProject
+      scProjectPhase
       scDescriptionEn {
         json
       }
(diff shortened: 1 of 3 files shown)


shawn320 (approved):
Nice work! I'm glad there is a way to filter content on graphql, thank you for doing this.

PR #648, opened 2023-01-23
update va update page and breadcrumbs to use data from AEM

shawn320:
# Description

Migrated virtual assistant post page to the new model. The page was using `Blog-v1` model, switching to the `Page-v1` model.

I also updated the breadcrumb items for each page to use data from AEM

## Test Instructions

1. Check the virtual assistant post page projects/virtual-assistant/what-we-learned, make sure the content are correct, and the links work properly
2. Check the breadcrumb links for each pages

### DIFF
--- a/__mocks__/mockStore.js
+++ b/__mocks__/mockStore.js
@@ -580,6 +580,21 @@ export const signupPage = {
         scDescriptionFr: {
           json: null,
         },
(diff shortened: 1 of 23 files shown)


will0684 (approved):
Tested and seems to be working as expected across the queries. Nice work thank you!

PR #650, opened 2023-01-31
Update Projects Cards

will0684:
# Description

Updated project page to support new thumbnail images on project cards served from AEM along with some slight margin changes. 

~~You might wonder why I used Next/Image instead of ``. Long story short I simply could not get images to load using the img tag inside the Card component. Through some reading I found that the issue of long load times with Next/Image was likely due to to not indicating a sizes property to control the size of images fetched for a given viewport. You can read about that [here](https://nextjs.org/docs/api-reference/next/image#sizes).~~

Images now use `` tag.

## Acceptance Criteria

Images should render on project cards and look reasonably good across all viewport sizes.

## Test Instructions

1. Navigate to projects page
2. See images on projects cards
3. Try different viewport sizes

## Checklist

- [x] Update CHANGELOG


shawn320 (approved):
Thanks!

PR #662, opened 2023-02-15
handle period at end of Alert/disclaimer link as separate data node

will0684:
# Description

Adds line to handle the period at the end of the homepage Alert component so that it is not a part of the link.

## Acceptance Criteria

Period should be separate to the link in the Alert/disclaimer on the home page. 

## Test Instructions

1. Got o homepage
2. Scroll down to the Alert
3. See that the period is not a character in the link

### DIFF
--- a/pages/home.js
+++ b/pages/home.js
@@ -246,6 +246,7 @@ export default function Home(props) {
                         .value
                     }
                   </a>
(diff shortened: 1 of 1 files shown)


andr0272 (approved):
:+1: Looks good!

PR #668, opened 2023-02-16
Get project stage from page properties and compose project stage tooltip from tooltip content fragment

will0684:
# Description

Previously, the "Information" label content was hardcoded into the component. Now it is fetched from the scTitle in the tooltip fragment attached to the OAS page model so it can be modified from AEM.

Also, project stage on the OAS page is pulled from the scLabProjectStage property found in the page model's properties. The tooltip content is pulled from the tooltip content fragment associated with the OAS page like the "Information" label content.

Screenshot 2023-02-16 at 10 19 52 AM

## Acceptance Criteria

Project stage should show Alpha, and the tooltip should correspond in both languages.

## Test Instructions

1. Go to OAS overview page
2. Ensure label says Alpha and the tooltip works and corresponds


shawn320 (approved):
Thank you!

PR #669, opened 2023-02-16
update yarn.lock

will0684:
# Description

Snyk package updates seem to be breaking the builds in the pipeline due to yarn.lock not being updated before it runs with --frozen-lockfile in the pipeline. Right now this means having to bring in the branch with the snyk update to your local machine, running `yarn install` to update the yarn.lock and then pushing out that change. We should explore our options to automate this better.


andr0272 (approved):
:yarn: :angry:

PR #690, opened 2023-03-01
ActionButton in flex container to compute same content size as other buttons on EE page

will0684:
# Description

The second "Give feedback" ActionButton on the EE page had a smaller computed content area then the first two ActionButtons on the page despite the same classes and fontsize. This was due to the flex container likely adding some rules for how items are vertically spaced. Solution here is to put the second button in it's own flex container. 

## Acceptance Criteria

Second "Give feedback" ActionButton should have the same layout as the first "Give feedback" ActionButton.

## Test Instructions

1. Inspect both buttons
2. Go to computed tab
3. See that they have the same padding and content area


shawn320 (approved):
Nice work, thanks!

PR #695, opened 2023-03-14
Incremental Static Regeneration

will0684:
# Description

This PR is to test out ISR implementation.

Removed the caching in AEMService because it would require us to call aemServiceInstance.flush() in getStaticProps before each request. Additionally, we only make requests once at build-time and all subsequent revalidation requests should not be fetching from cache and instead from the AEM resource for updated content. If this works as-is in dev I will add the revalidate prop to the rest of the pages.

## Acceptance Criteria

If you'd like to see ISR working, you can `yarn build` and then `yarn start` the application locally, then make changes to the OAS Benefits Estimator overview page content within AEM (make sure to do this after running the build). Refreshing the OAS page should show your changes after approximately 10 seconds. Don't forget to revert the content in AEM to their original values.

## Test Instructions

See above.

### DIFF
--- a/pages/about.js
+++ b/pages/about.js
@@ -177,5 +177,6 @@ export const getStaticProps = async ({ locale }) => {
       adobeAnalyticsUrl: process.env.ADOBE_ANALYTICS_URL,
       ...(await serverSideTranslations(locale, ["common"])),
     },
+    revalidate: 10,
   };
 };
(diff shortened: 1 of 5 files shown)


andr0272 (approved):
Saw it live, looks great. Keep an eye on that commented code :+1:

PR #700, opened 2023-03-21
update content to use data from AEM

shawn320:
# Description

[Go through application and ensure all strings (if possible) are coming from the CMS](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities/?workitem=111715)

Went through the app noticed the meta data for the home page still comes from local resource files, and part the meta data for the EE page is incorrect and using placeholder. I updated and published the meta data section for both page on AEM and update the app to use those data.


will0684 (approved):
Very nice 👍

PR #725, opened 2023-04-14
Addressing test warnings

will0684:
# Description

This PR modifies a number of tests that while passing, were throwing warnings related to:

- actions that update state not being wrapped in act()
- navigation not being implemented

Rather than suppressing the warnings, the tests have been configured as recommended in React 18 and the updated react testing library. The only exception to this is `Layout.test.js` where due to next/link, state changes occur outside of act() regardless of what you wrap with it.

## Acceptance Criteria

Running `yarn test:unit` should show tests passing with no warnings.

## Test Instructions

1. `yarn install`
2. `yarn test:unit`
3. Tests should pass without errors and without console warnings


andr0272 (approved):
Looks good. Just noting here, we get the error `Jest has detected the following 7 open handles potentially keeping Jest from exiting` when testing. We will create a new task to address this.

PR #726, opened 2023-04-17
use font-body class on list items in oas page

will0684:
# Description

List items on the OAS page are 16px, rather than 20px which is the proper font-size for body content. This PR adds the `font-body` class to the list items to make the font-size 20px.

## Acceptance Criteria

List items on OAS page should be 20px at all viewport sizes

## Test Instructions

1. Navigate to OAS page
2. Inspect list items to see 20px font-size


andr0272 (approved):
Perfect. Is it possible to do this from the <ul> element versus adding a <p> element to each list item?

andr0272 (approved):
Even better :+1:

PR #749, opened 2023-05-04
Remove nested anchor from Card.js

will0684:
# Description

Previous change to the Card.js component to allow right clicking to open a new tab across the component created an error with nested `<a>` tags. Essentially it's bad html to have an `<a>` tag nested within another `<a>`. This PR fixes the issue by replacing the unnecessary nested `<a>` tag with a `<p>` tag. Behaviour and appearance remains the same.

## Acceptance Criteria

Card should appear and behave the same way as before, without errors of bad html formatting in the console.

## Test Instructions

1. Navigate to the home page
2. Check the console for errors related to nested `<a>` tags. There should be none.
3. Right-clicking anywhere on the component should let you open a new tab

## Checklist

- [x] Update CHANGELOG


andr0272 (approved):
Nice catch :+1:

PR #763, opened 2023-05-26
Add adobe to csp whitelist

will0684:
# Add adobe to more areas of CSP

Added *.adobe.com to connect-src, script-src and frame-src.


shewood (approved):
looks good.

PR #764, opened 2023-05-26
Adobe CSP re-add *.adobe.com

will0684:
Adding *.adobe.com to CSP except `script-src`, as it gives warnings:

Screenshot 2023-05-26 at 11 04 36 AM


shewood (approved):
The wildcard seems like a bad idea but apparently AA said this is what you need to do.

PR #816, opened 2023-07-25
Set font-size in footer to 14px

will0684:
# Set font-size in footer to 14px

Default list-item font-size was set to 20px to accommodate the majority of lists that are included in body content. This made the footer list-items also 20px when they should be 14px per the design. This PR fixes this issue by adding a class to the list-items in the footer.

## Test Instructions

1. Go to home page, see that footer font-size is 14px
2. Repeat for other pages

### DIFF
--- a/components/organisms/Footer.js
+++ b/components/organisms/Footer.js
@@ -23,7 +23,7 @@ export function Footer(props) {
               return (
                 <li
                   key={index}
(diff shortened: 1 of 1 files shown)


andr0272:
Check out a preview of your pull request here: https://alphasite-appservice-dev-pr-preview-1448968039.azurewebsites.net

andr0272 (approved):
Sweet :+1:

PR #823, opened 2023-07-27
Increase width of Footer text content to avoid wrapping in french

will0684:
# Increase width of Footer text content to avoid wrapping in french

This was an issue noted by Stephanie where footer text when in French would wrap to a second line on smaller displays unnecessarily.

Screenshot 2023-07-27 at 11 01 34 AM

Here it is with the change:

Screenshot 2023-07-27 at 11 01 47 AM

This is a temporary fix until SCDS releases their new Footer component to production.

## Test Instructions

1. Go to homepage
2. Reduce display size to 550px
3. See that french text does not wrap to a second line

### DIFF
--- a/components/organisms/Footer.js
+++ b/components/organisms/Footer.js
@@ -23,7 +23,7 @@ export function Footer(props) {
               return (
                 <li
                   key={index}
-                  className="text-xs text-white w-64 md:w-56 lg:w-80 my-2.5 hover:underline list-none"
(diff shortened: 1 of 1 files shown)


andr0272:
Check out a preview of your pull request here: https://alphasite-appservice-dev-pr-preview-1452478250.azurewebsites.net

andr0272 (approved):
looks good :+1:

PR #825, opened 2023-08-02
Update README.md

will0684:
# Update README

Updated README to align more closely with how we set things up in our day-to-day dev and removed instructions for features we no longer leverage.

Please proofread + contribute anything we should have that is missing.


andr0272:
Check out a preview of your pull request here: https://alphasite-appservice-dev-pr-preview-1460169571.azurewebsites.net

Fbasham (approved):
Eh, I think this is good to ship.  I actually ran the project successfully without yarn, instead using npm, and it worked well enough.  I had issues with the env vars, but after asking you it seems to have worked out fine.  Ultimately new devs onboarding will likely encounter issues in any project, but this is a decent start for any new dev.

PR #838, opened 2023-08-15
Revert dependencies to v1.9.1

will0684:
# Revert dependencies to v1.9.1

While testing out pages as part of other work I noticed that on `main`, navigation to the OAS update page from the card on the OAS project page would fail both with yarn dev and with yarn build but works after a refresh or second attempt to navigate to the page. Additionally, when on the update page, switching from French to English would crash the page.

Screenshot 2023-08-15 at 11 06 01 AM

The error **seems** simple enough but after time spent testing, the issue appearss to be rooted in a dependency change brought in as part of https://github.com/DTS-STN/Service-Canada-Labs/pull/831, which included many dependency changes.

For now I am reverting most of the dependencies to what they were in v1.9.1. We may want to investigate implementing an e2e tests for navigation paths to the update pages which would at least help us in detecting these kinds of issues without having to manually test for this in the future.

## Test Instructions

1. Go to `main` branch
2. yarn install
3. Navigate to OAS project page
4. Scroll to and navigate to the update page through the update card
5. See that navigation fails
6. Refresh the page
7. See that page renders
8. Switch to French and then back to English
9. See that page crashes

--------------------------------------------------------
1. Bring in this branch `revert-deps-v1.9.1`
2. yarn install
3. Navigate to OAS project page
4. Scroll to and navigate to the update page through the update card
5. See that navigation works
8. Switch to French and then back to English
9. See that navigation works


Fbasham (approved):
Oops.  Hmm, definitely will need some investigation.  Clearly we have some dependency issues somewhere.  Will be fun to untangle this mess ;)

will0684:
> Oops. Hmm, definitely will need some investigation. Clearly we have some dependency issues somewhere. Will be fun to untangle this mess ;)

Yes the tree is...not optimal. Something for us to pick apart in the future (or +1 reason for a fresh repo TBD).

PR #858, opened 2023-08-29
[fix] Query for article pages directly instead of as a property of the parent project page

will0684:
# [Query for update pages directly (instead of as a property of the parent project page) to fix issue with article pages not showing breadcrumbs for project pages](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities/?workitem=135491)

This fixes an issue where breadcrumbs on article pages were not displaying the link to the preceding project page. This arose from an issue with AEM queries where, when querying for article pages as an associated property of the project page, the breadcrumbs for that associated article would unfortunately not include the breadcrumb of the parent project page. To fix this, I replaced the queries on the dynamic pages to query for their respective articles "directly".

## Test Instructions

1. Navigate to either OAS or BN article pages
2. See that breadcrumb includes the preceding project page

### DIFF
--- a/graphql/queries/benefitsNavigatorArticlesQuery.graphql
+++ b/graphql/queries/benefitsNavigatorArticlesQuery.graphql
@@ -1,14 +1,6 @@
 query getBenefitsNavigatorArticles {
   scLabsPagev1List(
     filter: {
-      scContentType: {
-        _expressions: [
-          {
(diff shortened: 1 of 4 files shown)


Fbasham (approved):
I still have no idea how to effectively write GQL ,but this does the trick.  Good work right here, thank you.

PR #861, opened 2023-08-30
add svg chevron icon to breadcrumb

Fbasham:
# [Chevron icon for breadcrumbs is not appearing in local dev](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities/?workitem=138367)

I introduced this bug by removing icomoon.  Totally my fault.  Here's a fix that instead uses fontawesome 
(we're already depending on this elsewhere in the project) in the breadcrumb component:

image


will0684 (approved):
Looks good thank you!

PR #862, opened 2023-08-31
change csp and how fonts are loaded

Fbasham:
# [Fix issue with font variations not loading due to CSP config](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities/?workitem=138549)

- Move font import into `globals.css`
- Modify CSP in `next.config.js`

Hopefully now there isn't an error in the console about failing load fonts because the CSP.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -15,19 +15,19 @@ const REWRITES = [
   },
   {
     source: "/projets/estimateur-prestations-sv/:slug",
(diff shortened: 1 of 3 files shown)


will0684 (approved):
Awesome thank you!

PR #865, opened 2023-09-01
remove next-page-tester

Fbasham:
# [Investigate if we still need to leverage next-page-tester](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities)


will0684 (approved):
Tested and working well thank you!

PR #867, opened 2023-09-11
[fix] Modify dcterms.title and dcterms.creator values

will0684:
# [Modify dcterms.title and dcterms.creator values](https://dev.azure.com/VP-BD/DECD/_boards/board/t/Service%20Canada%20Labs/Stories%20and%20Activities/?workitem=140509)

This addresses issues raised by the AA Data team related to null value page titles in AA and to an incorrect value of "Service Canada" for dcterms.creator. Task card in ADO includes the change request email.

### DIFF
--- a/pages/404.js
+++ b/pages/404.js
@@ -72,7 +72,14 @@ export default function error404(props) {
             title="gccore"
             content={t("metaSubject")}
           />
-          <meta name="dcterms.creator" content="Service Canada" />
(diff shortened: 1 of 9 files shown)


Fbasham (approved):
up to you if you want to move these into the common locale files since `t` is already being used.  It's a small nitpick, but either way I'm fine with this

PR #886, opened 2023-09-26
update ZAP target

Fbasham:
ZAP scans have been failing silently since the target URL has changed.  To see this, go back to previous commits and inspect the ZAP scan actions.  The scans have 502ing and silently failing, yet seem to pass the check.  I'm not sure, but hopefully this PR fixes the issue.


andr0272:
Check out a preview of your pull request here: https://alphasite-appservice-dev-pr-preview-2.azurewebsites.net

PR #887, opened 2023-09-26
update ZAP

Fbasham:
Modify ZAP test to be more in line with other DTS products.  Before this PR, ZAP was being run against every pull request against the (now) outdated alpha site URL.  Because the URL was outdated, the test didn't actually do anything and would silently fail (this can be seen if you check the logs on previous PRs).  This change aligns SC Labs with other DTS products by running the ZAP test against the main branch after a merge.


andr0272:
Check out a preview of your pull request here: https://alphasite-appservice-dev-pr-preview-1.azurewebsites.net

andr0272:
Check out a preview of your pull request here: https://alphasite-appservice-dev-pr-preview-1.azurewebsites.net

PR #892, opened 2023-09-28
remove oas ee new tag and update logic for displaying tag

Fbasham:
# [Implement "New update" tag guidance to the home page](https://dev.azure.com/VP-BD/DECD/_workitems/edit/140804)

The content chapter, design team, and devs have agreed that the approach to take with regards to displaying the "New Update" tag on project cards is to have a field in AEM named `scLabNewExpiryDate`.  This is the date (not inclusive) at which the new tag should *not* be displayed on project cards on the home page.  

### Changes
- update gql query to pull in the `scLabsNewExpiryDate` on the project query
- add conditional to only render the tag if the current datetime is less than or equal to the expiry date in AEM.

### DIFF
--- a/graphql/queries/projectQuery.graphql
+++ b/graphql/queries/projectQuery.graphql
@@ -18,6 +18,7 @@ query getAllProjects {
       scTitleFr
       scPageNameEn
       scPageNameFr
+      scLabsNewExpiryDate
       scBreadcrumbParentPages {
         scTitleEn
         scTitleFr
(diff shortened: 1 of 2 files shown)


andr0272:
Check out a preview of your pull request here: https://alphasite-appservice-dev-pr-preview-0.azurewebsites.net

will0684 (approved):
Looking good thank you!

PR #904, opened 2023-10-18
[fix] ol and ul should have same layout style

will0684:
Small fix to match layout style of unordered lists and ordered lists


Fbasham (approved):
I concur.

PR #925, opened 2023-11-15
Set node version to 18.x in Github workflows

will0684:
This PR sets the node version in our github workflow actions to at least version 18. Reason to do this is that Github actions use Node v16 by default, but because we are using Yarn v4, we want to ensure yarn installs are running in an environment using at minimum Node v18.


andr0272:
Check out a preview of your pull request here: https://app-servicecanadalabs-dev-pr-preview-3.azurewebsites.net

Fbasham (approved):

PR #926, opened 2023-11-15
Remove install-state.gz from tracking

will0684:
According to [this](https://yarnpkg.com/getting-started/qa), we can remove this file safely:

"`.yarn/install-state.gz` is an optimization file that you shouldn't ever have to commit. It simply stores the exact state of your project so that the next commands can boot without having to resolve your workspaces all over again."


andr0272:
Check out a preview of your pull request here: https://app-servicecanadalabs-dev-pr-preview-1.azurewebsites.net

Fbasham (approved):

PR #1025, opened 2024-05-15
Upgrade Nextjs version to fix issue with sharp and output:standalone

will0684:
# Upgrade Next version to fix issue with `sharp` and `output:standalone`

This PR upgrades Nextjs to version 14.2.3 from our previous version 14.0.1.

The reason for this upgrade is that because we use `output:standalone` in `next.config.js`, we had problems with the production image being able to resolve the `sharp` dependency for optimizing images. [This was a known issue](https://github.com/vercel/next.js/discussions/40125).

Now images should have faster optimization in both production and development environments as both environments run a production build with `output:standalone`.

## Test Instructions

1. yarn build + yarn start
2. Navigate the site as much as possible and ensure images are loading quickly and without errors


andr0272:

PR #1140, opened 2024-12-31
Dependency Reversion to state at v2.0.1

will0684:
The purpose of this PR is to revert dependencies to how they were at v2.0.1 to fix an issue with image optimization using `sharp` failing the upstream fetch when running dev and prod builds.

I will be making unit tests for the Image component to test that optimization is working as expected and have created a card to do so.

## Test Instructions

1. yarn dev
2. navigate to /home
3. see that images are loading as expected, without fetch errors for the images in the console
4. repeat using yarn build + yarn start (ignore "can't use yarn start" error)


andr0272:

PR #1144, opened 2025-01-08
use unoptimized={true} on next/image

will0684:
Image optimization is failing consistently and returning 500 errors from the server. There seems to be an issue in the dependency tree but after some testing the issue has been difficult to pinpoint.

For the time being, image optimization will be deactivated until more investigation can be made into the issue.

## Test Instructions

1. Navigate to /en/home
2. See that images render properly
3. Navigate to /en/projects
4. See that images render properly


andr0272:

PR #1215, opened 2025-03-14
Upgrading OWASP ZAP version to 0.12.0 and checkout to v4

lhr-git:
# Updating GH Workflow versions (OWASP ZAP and checkout)

As part of fixing the OWASP ZAP action, the docker pull request on the ZAP action isn't functioning. While looking into this I noticed that the versions were out of date as part of this action, this is just a patch to update them to the current version.

## Test Instructions

Unfortunately, none, this is a github specific workflow update.

## Definition of Done

- The workflow action versions have been updated to 0.12.0 (OWASP ZAP) and v4(checkout)

### DIFF
--- a/.github/workflows/dast.yml
+++ b/.github/workflows/dast.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


will0684 (approved):
Thank you!

PR #1226, opened 2025-03-25
fix: Update TextContentMapper to handle default layout case

will0684:
* Modified TextContentMapper to return a default layout when no specific layout is found for the fragment data. This change ensures that the component can gracefully handle undefined layout configurations.


andr0272:
Check out a preview of your pull request here: https://app-servicecanadalabs-dev-pr-preview-3.azurewebsites.net

